PR #15, opened 2017-10-04
add cmake build system support

lrusak:
Here it is, was quite a bit simpler than libnfs

I've only tested this on linux (needs testing on windows, osx, solaris, and whatever aros is)

It contains all the functionality that autotools did.

You can now use the following build options
```
BUILD_SHARED_LIBS
ENABLE_EXAMPLES
```

Also, this installs the following libraries
```
libsmb2.so -> libsmb2.so.1
libsmb2.so.1 -> libsmb2.so.1.0.0
libsmb2.so.1.0.0
```

You can build as follows
```
mkdir -p build
cd build
cmake -DBUILD_SHARED_LIBS=ON
      -DCMAKE_INSTALL_PREFIX=/usr/local
      -DENABLE_EXAMPLES=ON
      ..
make -j9
make test
```

Please let me know if you have any questions or concerns


sahlberg:
Merged. Thanks.

AROS is basically a re-implementation of Amiga OS 1.3 (home computer in the late 80s).
I don't think they have C-make :-)
(they barely have basic berkely socket support :-)

Issue #16, opened 2018-03-13
Some hosts send wrong compound error response.


afedchin:
Looks like this line https://github.com/sahlberg/libsmb2/blob/master/lib/smb2-cmd-error.c#L77 causes an overlapping the body with next header in handling compounded response with errors. while the expected size of response is 9 it actual size is 8. this causes an overlapping the body to next header. this makes unimpossible to use the same context to any other requests.

![image](https://user-images.githubusercontent.com/2063885/37328179-3c0f2368-26aa-11e8-9eaf-bec2238a3b74.png)

if I comment this line all works fine

sahlberg:
The error packet is generated by the server, not libsmb2.
Even wireshark dissects this packet wrong.

Which server is this that sends broken error packets?
Just for curiosity. We still need to change libsmb2 to work with this broken servers as well.

afedchin:
Windows 10 Enterprise FCU (10.0.16299.309)

afedchin:
Yes [MS-SMB2] is saying the next:
> 2.2.2 SMB2 ERROR Response
>  If the ByteCount field is zero then the server MUST supply an ErrorData field that is one byte in length, and SHOULD set that byte to zero; the client MUST ignore it on receipt.<4>

The key words here are `the client MUST ignore it on receipt.` 
If we take a look at `NextCommand`, it has correct value = 0x48. 72-64 (header) = 8 i.e. even if error response must be 9+1 (iiuc) length the server makes it smaller assuming that the client should ignore it.

my "fix" looks correct if we just drop `if` statement [here](https://github.com/sahlberg/libsmb2/blob/master/lib/smb2-cmd-error.c#L76) and always return `ByteCount` value even if it zeroed. In any case this response has no data and has no any effect if we just ignore it as the doc said.

sahlberg:
On Thu, Mar 15, 2018 at 5:48 PM, Anton Fedchin <notifications@github.com>
wrote:

> Yes [MS-SMB2] is saying the next:
>
> 2.2.2 SMB2 ERROR Response
> If the ByteCount field is zero then the server MUST supply an ErrorData
> field that is one byte in length, and SHOULD set that byte to zero; the
> client MUST ignore it on receipt.<4>
>
> The key words here are the client MUST ignore it on receipt.
> If we take a look at NextCommand, it has correct value = 0x48. 72-64
> (header) = 8 i.e. even if error response must be 9+1 (iiuc) length the
[...]

PR #19, opened 2018-03-19
A set of fixes

afedchin:
this fixes some issues including #16 
adds build with msvc

tested with Kodi uwp

@sahlberg please check that this doesn't break build for posix platforms


sahlberg:
Merged, thanks.
Minor posix+krb5 fixes required, no drama.

Thanks!

PR #22, opened 2018-03-29
Use krb5 credential cache to get and store the TGT and the creds.

skbehera1984:
Use krb5 credential cache to get and store the TGT and the creds. So instead of calling gss_acquire_cred we can use gss_krb5_import_cred which will use
the cached creds for the user. We use here an IN-MEMORY cache.

This will help user avoid the kinit command before using this library. and also no need of adding details in krb5.conf.

### DIFF
--- a/lib/krb5-wrapper.c
+++ b/lib/krb5-wrapper.c
@@ -53,6 +53,8 @@
 #include <unistd.h>
 #endif
 
+#include <krb5/krb5.h>
+#include <gssapi/gssapi_krb5.h>
 #include <gssapi/gssapi.h>
 #include <stdio.h>
 
@@ -63,6 +65,13 @@
 #include "libsmb2-private.h"
 
 #include "krb5-wrapper.h"
+
+/** Global Definitions */
+krb5_context    krb5_cctx;
+krb5_ccache     krb5_Ccache;
+krb5_creds      krb5_ccreds;
+krb5_principal  client_princ;
+
 
 void
 krb5_free_auth_data(struct private_auth_data *auth)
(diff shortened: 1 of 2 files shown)


lib/krb5-wrapper.c
```diff
@@ -137,14 +146,127 @@ krb5_set_gss_error(struct smb2_context *smb2, char *func,
...
+    ret = krb5_cc_new_unique(krb5_cctx, "MEMORY", NULL, &krb5_Ccache);
+    if (ret != 0)
+    {
+      smb2_set_error(smb2, "Failed to create krb5 credentials cache - %s", krb5_get_error_message(krb5_cctx, ret));
+      return -1;
+    }
+
+    len = strlen(smb2->domain);
+    ret = krb5_build_principal(krb5_cctx, &client_princ, len, smb2->domain, user, NULL);
+    //ret = krb5_parse_name(krb5_cctx, user, &client_princ);
```
sahlberg: dead code ?

lib/krb5-wrapper.c
```diff
@@ -137,14 +146,127 @@ krb5_set_gss_error(struct smb2_context *smb2, char *func,
...
+    if (client_princ != NULL)
+      krb5_free_principal(krb5_cctx, client_princ);
+
+    krb5_free_cred_contents(krb5_cctx, &krb5_ccreds);
+    krb5_cc_destroy(krb5_cctx, krb5_Ccache);
+    krb5_free_context(krb5_cctx);
+    return 0;
+}
+
+// Enable thhis flag to use krb5 cached creds or else you need to do a kinit
```
sahlberg: no c++ comment

PR #23, opened 2018-03-29
implement rename

afedchin:
@sahlberg ping

### DIFF
--- a/lib/smb2-cmd-set-info.c
+++ b/lib/smb2-cmd-set-info.c
@@ -55,6 +55,7 @@ smb2_encode_set_info_request(struct smb2_context *smb2,
         uint8_t *buf;
         struct smb2_iovec *iov;
         struct smb2_file_end_of_file_info *eofi;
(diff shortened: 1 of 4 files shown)


lib/smb2-cmd-set-info.c
```diff
@@ -92,6 +93,35 @@ smb2_encode_set_info_request(struct smb2_context *smb2,
...
+                        }
+
+                        len = 20 + name->len * 2;
+                        smb2_set_uint32(iov, 4, len); /* buffer length */
+
+                        buf = malloc(len);
+                        if (buf == NULL) {
+                                smb2_set_error(smb2, "Failed to allocate set "
+                                               "info data buffer");
+                                return -1;
```
sahlberg: free(name)) here ?

afedchin: my fault. will fix

sahlberg: I have merged it. I will fix it in a small followup patch

PR #24, opened 2018-03-30
Use krb5 credential cache to get and store the TGT and the creds.

skbehera1984:
Use krb5 credential cache to get and store the TGT and the creds. 
So instead of calling gss_acquire_cred we can use gss_krb5_import_cred which will use
the cached creds for the user. We use here an IN-MEMORY cache.

This will help user avoid the kinit command before using this library. and also no need of adding details in krb5.conf.

### DIFF
--- a/README
+++ b/README
@@ -100,14 +100,12 @@ KRB5 Authentication
 Kerberos authentication can be used when the linux workstation as well as
 the file server are part of Active Directory.
 
-Before you can use KRB5 authentication you must first authenticate to the KDC.
-Use "kinit username@YOUR-REALM" to do that in the normal way you set up
-kerberos.
-
-Once you have run kinit successfully you should be able to authenticate to the
-file server using krb5 by specifying sec=krb5 in the URL :
+You should be able to authenticate to the file server using krb5
+by specifying sec=krb5 in the URL :
   smb://server/share?sec=krb5
 
+The application needs to set the username, password and the domain fqdn in the context using
+smb2_set_user(), smb2_set_password() and smb2_set_domain() respectively.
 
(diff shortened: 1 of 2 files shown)


sahlberg:
Merged, thanks

sahlberg:
Using #define USE_CACHED_CREDS to decide whether to use kinit or cached credentials at compile time is perhaps not great.

This would be better to be able to set dynamically at runtime.
For example using  "sec=krb5cc" in the url to specify we want cached creds instead of kinit.

What do you think? Can you send a PR with such change?

sahlberg:
I have changed the patch so that instead of USE_CRED_CACHE at compile time, now you select
cread cache by setting sec=krb5cc in the URL.
If you set sec=krb5 then you get the initial behaviour that depends on an initial kinit.

Can you please have a look and test that your credential cache still works ?

skbehera1984:
Thanks Ronnie, it is working fine.

sahlberg:
On Sun, Apr 1, 2018 at 5:17 PM, skbehera1984 <notifications@github.com>
wrote:

> Thanks Ronnie, it is working fine.
>
>
Thanks for testing and confirming it works.
I am always curious about the use case for people using or contributing to
my projects.
If nothing else it makes me better understand the end-user needs for the
library.
(The end-user in this case is the developer that consumes the library and
uses it is his/her application.)

If you can share with me your use case that would be awesome.
My email is ronniesahlberg@gmail.com
If you can not share it, that is fine too. Lots of people can not, or do
not want to share that and that is fine.


Thanks for using libsmb2
Thanks for contributing and making it better!

regards
ronnie sahlberg

PR #26, opened 2018-03-30
Add sync API smb2_echo()

skbehera1984:


afedchin:
my fault, forgot about sync API :)

sahlberg:
Merged, thanks.

(afedchin, that is ok, we want to discourage use of the sync api)

PR #27, opened 2018-04-05
Avoid using naked krb5 calls to use cached creds instead use gssapi calls

skbehera1984:
Testes with krb5 using kinit and with krb5cc without kinit


skbehera1984:
Enabled SMB2 signing for krb5 sessions.

sahlberg:
Some quick comments, Very nice !


There were a few // style comments. Please update the patches and change
them to /* */ style comments.

I think for signing we want to think about what the semantics should be.

I think, default should be to use signing if the server supports it and
fallback to no signing if not.
On top of this we should have a new url argument :
"sign" : which means must use signing and fail the connect if the server
does not support it.


Do you think you can do those changes ?


regards
ronnie sahlberg

skbehera1984:
I have changed the comments to C style. Sorry have a habit of using c++
comments.

For the other queries, I think the server doesn't leave an option for us.
If the server requires signing then we must sign the pdu. If the server
only has signing enabled but has not made it mandatory then it is our
choice.

Thanks
-Sarat

sahlberg:
On Tue, Apr 17, 2018 at 6:54 PM, skbehera1984 <notifications@github.com>
wrote:

> I have changed the comments to C style. Sorry have a habit of using c++
> comments.
>
> For the other queries, I think the server doesn't leave an option for us.
> If the server requires signing then we must sign the pdu. If the server
> only has signing enabled but has not made it mandatory then it is our
> choice.
>

What I mean is that it would be nice to have a way for the client to
require that
signing must be used and fail the connection if not.

I.e. in session setup, set SecurityMode.SMB2_NEGOTIATE_SIGNING_REQUIRED


regards
ronnie sahlberg

skbehera1984:
Yes we can do that. I can do that in a separate change.

sahlberg:
On Tue, Apr 17, 2018 at 7:13 PM, skbehera1984 <notifications@github.com>
wrote:

> Yes we can do that. I can do that in a separate change.
>

Cool, lets do that as a separate patch.
I will review this tomorrow and merge unless there is any major issues.

afedchin:
Is it possible to use signing with internal NTLMSSP?

lib/smb2-signing.c
```diff
@@ -0,0 +1,204 @@
...
+#include <unistd.h>
+#endif
+
+#include "smb2-signing.h"
+
+#include <openssl/hmac.h>
+#include <openssl/cmac.h>
+#include <openssl/evp.h>
+#include <openssl/sha.h>
+
```
sahlberg: It adds a new dependency on openssl.
Can you add configure checks if openssl is present and conditionally remove support for signing if not ?

stefbon: I see that openssl is used. It's also possible to use libgcrypt. Are there special reasons to use openssl?
Futher I see that with the function CMAC_AES_128 for example the context for creating the cmac is created and freed for every message. It's also possible (since the algorithm used does not change during runtime) allocate them once and reuse these ctx's. I cannot oversee one ctx is enough or more are required.
Stef Bon

skbehera1984: I am not good at the automake tools, can you help me in this regard please.

sahlberg:
No problem.   I will add the configure changes.

skbehera1984:
I have done the changes, please check if any modifications are required. Thanks

sahlberg:
I really want this in libsmb2 but it needs some work.

First we need to split this into two different pull requests, one for the first two patches that operate on the creds cahce, and a separate one for the signing.

￼
First pull request:
=============
The first pull request should contain these three patches:
Drop using the naked krb5 APIs to manage the krb5 creds cache. ￼…4a11175
Getback the krb5cc flag and free up the domain and workstation memory   9224132
GSSAPI uses the default creds cache. Re-direct the cache to MEMORY … 2c54ea5

But it is messy that the first patch removes a lot of stuff and the second adds the exact same thing back.
Here I think it is better if these two are just squashed into a single patch.

To do that, create a new branch based on my master then git cherry-pick these two patches into it.
To squash them into a single commit you can then git rebase -i HEAD~3 and "fold" the last patch into the previous patch.
That turns it into a single patch.

Now, there is still a problem with the patch. It starts to use smb2->password and dereference it.
This variable is only set programatically (via smb2_set_password() which is only supposed to be used for applications that use/link against the builtin ntlmssp provider. I.e. kodi/vlc/...

Normal users on linux/unix would use the shared library built against gssapi and mit kerberos and in those two cases smb2->password will pretty much always be NULL since either we use kerberos and there is no password, or we use the ntlmssp plugin mech in which case the password is in the NTLM_USER_FILE.

I.e.   all the examples such as ./examples/smb2-ls-sync etc will segfault with these patches since smb2->password is NULL.
So for this to work we would need a change that does not require smb2->password.
I am not familiar with the API you switched to enough to say if that is possible or not.


Second pull request:
================
The second pull request should contain the patches we need to add the signing.
I.e. the other patches.

First comment, If signing initially only works for smb2 but not 3, that is fine.

You saw I added a patch where you can specify the version to negotiate so that this will be possible to do more cleanly.
(I wasn't a fan of the smb2_enable_smb3() so I think this will be better. It will require some rebasing work for you. Sorry.)

So cherry pick these patches into a new branch and create a new pull request just for the signing.
Here there are a bunch of changes I want. The first and simplest change is style.
Indentation is 8 characters, not 4.
Braces for controlstatements go on the same line, i.e. use:
   if ... {
   } else {
   }
do not use
  if ...
  {
  }
  else
  {
  }

These are fairly trivial changes that only affect style and should be easy to do.
You don't have to create new patches to fix the formatting problems, you can change the patches themself (which is the preferred way in git) by using
git rebase -i HEAD~5   and then choose "edit" for the patch you want to edit and change.

git is really nice in this regard compared to a lot of other version control systems I have used.
I think this should be fairly trivial to fix.

The second problem with these patches are that they seem to use a deprecated openssl api
which means they do not build on modern fedora (and I suspect some others as well.)
It does compile on fedora 25, but If I try fedora 27, they use a different version of openssl and there
it will fail with 


smb2-signing.c: In function ‘HMAC_SHA_256’:
smb2-signing.c:110:14: error: storage size of ‘ctx’ isn’t known
     HMAC_CTX ctx;
              ^~~
smb2-signing.c:116:5: error: implicit declaration of function ‘HMAC_CTX_init’; did you mean ‘HMAC_CTX_new’? [-Werror=implicit-function-declaration]
     HMAC_CTX_init(&ctx);
     ^~~~~~~~~~~~~
     HMAC_CTX_new
smb2-signing.c:117:5: error: ‘HMAC_Init’ is deprecated [-Werror=deprecated-declarations]
     HMAC_Init(&ctx, key, key_len, EVP_sha256());
     ^~~~~~~~~

in fedora 26 openssl no longer has a HMAC_CTX structure nor a HMAC_INIT function.

Can you have a look ?

skbehera1984:
Sure. Let me try that.

PR #31, opened 2018-04-20
Drop using the naked krb5 APIs to manage the krb5 creds cache.

skbehera1984:
Instead use gss_acquire_cred_with_password() to create a temporary gss creds cache
and use it.

GSSAPI uses the default creds cache. Re-direct the cache to MEMORY
using krb5 APIs so that the TGT and creds will be destroyed once the
current session is complete.

### DIFF
--- a/README
+++ b/README
@@ -53,6 +53,9 @@ Aruments supported by libsmb2 are :
 		  2.02, 2.10, 3.00, 3.02 : negotiate a specific version.
 		  Default is to negotiate any SMB2 or SMB3 version.
 
+NOTE:-
+	When using krb5cc mode use smb2_set_domain() and smb2_set_password() in the examples and applications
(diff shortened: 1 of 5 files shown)


skbehera1984:
I have wrapped everything related to krb5cc mode inside use_cached_creds flag and added validations, so the examples can be used "AS IS" with krb5 mode.
To use the examples in krb5cc mode smb2_set_domain() and smb2_set_password() must be used in the examples.

sahlberg:
Thanks for doing these changes.

I have tested it and it works.
It looks good so I will merge with some minor changes (long lines) once I
get back from my travels.

sahlberg:
Merged, thanks!

sahlberg:
Merged now.

PR #32, opened 2018-04-21
use EBADF instead of EBADFD, missing on darwin

MrMC:
Replace EBADFD with EBADF Using EBADF is more portable and correct because EBADFD means a file descriptor in bad state, not an invalid file descriptor.

see https://mail.gnome.org/archives/commits-list/2015-December/msg01323.html


MrMC:
stupid github :) ignore the 2nd commit, see PR #34

sahlberg:
Merged, thanks !

PR #33, opened 2018-04-21
fix build for windows after b5ef492

afedchin:


afedchin:
looks like this repo needs a CI :)

sahlberg:
Fully agree I need CI here. One day.

Merged, thanks!

PR #37, opened 2018-04-28
Retrieve the SSPI session key after a successful SMB2 SessionSetup

skbehera1984:
This only works in LIBKRB5 mode.

This is a subset of pullrequest #27

### DIFF
--- a/include/libsmb2-private.h
+++ b/include/libsmb2-private.h
@@ -131,12 +131,14 @@ struct smb2_context {
         void *connect_data;
 
         int credits;
-        
+
         char client_guid[16];
-        
+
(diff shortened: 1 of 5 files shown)


sahlberg:
Thanks! 

Merged with some tiny style issues.
I changed a few lines of
if ...
{
into
if ... {

and there were some lines over 80 chars.

When you run 'git commit'  Please add the -s flag so you also get a signed-off-by line.
I added this manually.


Great work! Thanks  and I cant wait until we get singing working.

sahlberg:
Merged with some tiny style changes.

This is aswesome! We are are on the patch of getting signing support!

PR #38, opened 2018-05-02
Enable SMB2 signing :- Sign an outgoing pdu.

skbehera1984:
Use openssl libcrypto to compute the signatures
added config checks for openssl headers and library

SMB3 signing is not implemented yet.


sahlberg:
Nice.

There were a fe style issues, mainly
if
{
...


where I prefer the opening bracket on the same line, like this :
if {


That is minor and I can sort it out.


Testing this, it seems it only works for sec=krb5 while sec=ntlmssp fails.
Can you have a look and see if / what is wrong ?
(i.e. using ntlmssp with the kerberos libraries and the ntlmssp gssapi mech
plugin)

sahlberg:
In session_setup_cb() you need to close the context and invoke the user
callback with an error (-1)
or else we will hang indefinitely if krb_session_get_session_key() returns
an error.
Something like this :

#ifdef HAVE_LIBKRB5
        if (krb5_session_get_session_key(smb2, c_data->auth_data) < 0) {
+                smb2_close_context(smb2);
+                smb2_set_error(smb2, "Failed to get session key");
+                c_data->cb(smb2, -1, NULL, c_data->cb_data);
+                free_c_data(smb2, c_data);
                return;
        }
#endif

skbehera1984:
I am not using the plugin from "https://github.com/simo5/gss-ntlmssp". I am using a different library. And this code piece works for me. I just checked the plugin you mentioned, it doesn't allow inquiry of SSPI key. Code from there

    if (gss_oid_equal(desired_object, &spnego_req_mic_oid)) {
        return gssntlm_spnego_req_mic(minor_status, ctx, data_set);
    }
    return GSSERRS(ERR_NOTSUPPORTED, GSS_S_UNAVAILABLE);

That support needs to be added in the plugin. the current code will print "Failed to get session key" in ntlmssp mode.

skbehera1984:
Tested the following
-----------------------
1. Standalone ntlmssp mode
   a. if signing required by server, it fails with "smb2_connect_share failed. Signing Required by server. Not yet implemented"
   b. if signing not required by server, it works

2. libkrb5 mode + krb5
   a. if signing not required by server, it works.
   b. if signing required by server, it works

3. libkrb5 mode + ntlmssp
   a.if signing required by server, it fails with "smb2_connect_share failed. Signing required by server. Session Key is Not available"
   b.if signing not required by server, it works.

sahlberg:
On Thu, May 3, 2018 at 4:09 PM, skbehera1984 <notifications@github.com>
wrote:

> I am not using the plugin from "https://github.com/simo5/gss-ntlmssp". I
> am using a different library. And this code piece works for me. I just
> checked the plugin you mentioned, it doesn't allow inquiry of SSPI key.
> Code from there
>
> if (gss_oid_equal(desired_object, &spnego_req_mic_oid)) {
>     return gssntlm_spnego_req_mic(minor_status, ctx, data_set);
> }
> return GSSERRS(ERR_NOTSUPPORTED, GSS_S_UNAVAILABLE);
>
> That support needs to be added in the plugin. the current code will print
> "Failed to get session key" in ntlmssp mode.
>

I see. That is unfortunate :-(
Ok, I have tested your patch and looked over it. Aside from some style
issues I am happy with it.

I will go over it one more time and then merge tomorrow unless something
shows up.

sahlberg:
Merged.

Many thanks, this is awesome.
Do you want to try adding signing support to the builtin NTLMSSP module
too? That module is probably what most users will use as it avoids the
hassle of setting up krb5.conf

sahlberg:
Merged! Thanks, awesome work!

PR #40, opened 2018-05-11
SMB3 Signing Support:-

skbehera1984:
Derive the SMB3 signing key.
	Formulate the AES CMAC 128 algorithm
	Compute the MAC for SMB3.00 and SMB3.02

This doesn't work for SMB3.1. The key derivation requires
the preauth context to be extracted.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -186,7 +186,7 @@ AC_HAVE_LIBRARY(crypt,
 		], [])
 
 # check for openssl/hmac.h openssl/cmac.h openssl/evp.h openssl/sha.h
-AC_CHECK_HEADERS([openssl/hmac.h openssl/cmac.h openssl/evp.h openssl/sha.h openssl/opensslv.h],
+AC_CHECK_HEADERS([openssl/hmac.h openssl/aes.h openssl/evp.h openssl/sha.h openssl/opensslv.h],
                   [],
                   [AC_MSG_WARN([cannot find openssl headers. Building without SEAL support.])
 		  AC_DEFINE(HAVE_OPENSSL_LIBS, 0,
--- a/include/libsmb2-private.h
+++ b/include/libsmb2-private.h
@@ -27,6 +27,8 @@
 extern "C" {
 #endif
 
+#define MIN(a,b) (((a)<(b))?(a):(b))
+
 #ifndef discard_const
 #define discard_const(ptr) ((void *)((intptr_t)(ptr)))
 #endif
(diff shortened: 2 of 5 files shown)


sahlberg:
Merged, thanks!
Should have mentioned it in the review, but for the commit message, it should be formatted as
first one single line that describes the patch, then one blank line.

The blank line is important to git since git will use the first whole paragraph in the short listing.
You can see that effect if you run for example:
git rebase -i HEAD~1

which will print the name of the patch as : 
---
pick b094bc2 SMB3 Signing Support:-     Derive the SMB3 signing key.    Formulate the AES CMAC 128 algorithm    Compute the MAC for SMB3.00 and SMB3.02
---

No big deal.

Awesome to get smb3 signing support!
Did you think about adding signing support to the builtin ntlmssp provider too ?
That would be really nice since that one is what I suspect most users of kodi and friends would use.

sahlberg:
I have no idea why github made that text bold and a bigger size.

I did not intend to look like screaming

skbehera1984:
No worries. I didn't take that as screaming, but a guidance on the git
stuff. For the ntlmssp signing we just need the session key and the
remaining stuff are already in place. To get the session key we need to
understand the algorithm which is different for NTLM V1 and V2. I am not
aware of that by now. But will study that and give it a try.

Ideally it should be done through the gss APIs when ntlmssp is glued to the
gss module. But looks like the third party library that you used for
ntlmssp doesn't export the session key.

Having said that it is still possible to get the key in standalone mode.
Will try that.

Thanks
-Sarat

sahlberg:
There are multiple issues with the GSS module. It is a great module for linux and unixens that link with MIT and just want some NTLMSSP support.

This is the only open source or freely available GSS plugin I am aware of for NTLMSSP.
Do you know of any alternatives I could use instead?



Sorry but I have to enter rant-mode to explain:

However, the GSS module I link to is LGPLv3. Several consumers of the libsmb2 library can not use that licence. Some projects I care a lot about, like QEMU/KVM, are GPLv2-only and thus incompatible with LGPLv3.

Other projects, like media players want to provide this library with their app on various app stores. The problematic ones that comes to mind immediately are iOS and the windows UWP stuff. Both of which considered to be super incompatible with LGLv3, but are considered compatible with LGPLv2.1
Many of those users have to be extra-turbo-plus careful about licencing since, well becasue.
And a LGPLv2.1 library I think helps them.

These are big reasons why libsmb2 is LGPLv2.1 and not LGPLv3

I want as many people as possible, and as many apps as possible to have good SMB2 share support.

End rant mode.

PR #42, opened 2018-05-18
Signing support for the built-in NTLMSSP mode

skbehera1984:
For NTLMv2 the session key is to be calculated by the client from
the session base key, which inturn gets generated from the key exchange key.
Since in current implementation we are not doing
key exchange, the key_exchange key is both the session base key
and it is the session key.

Follow - https://blogs.msdn.microsoft.com/openspecification/2010/04/19/ntlm-keys-and-sundry-stuff/
for more details on NTLM keys

### DIFF
--- a/lib/libsmb2.c
+++ b/lib/libsmb2.c
@@ -602,42 +602,62 @@ session_setup_cb(struct smb2_context *smb2, int status,
                 free_c_data(smb2, c_data);
                 return;
         }
-#ifdef HAVE_LIBKRB5
-        if (smb2->signing_required &&
-            krb5_session_get_session_key(smb2, c_data->auth_data) < 0) {
-                smb2_close_context(smb2);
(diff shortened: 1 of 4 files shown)


afedchin:
@skbehera1984 can you please cherry-pick this commit https://github.com/afedchin/libsmb2/commit/d54556364b045a2d26e4ef270d68f5de4384cf87?

sahlberg:
Merged, thanks!

sahlberg:
@afedchin I cherry picked your patch into master after the merge.

Thanks.
What we need now I guess is to add NTLM_USER_FILE parsing for the builtin-ntlmssp so that the examples will work.

skbehera1984:
In addition to this you can combine builtin NTLMSSP with the libkrb5 meaning build both of them together. So there will be no need to depend on the gss-ntlmssp library. And then later you can add the gss mech-glue to the builtin ntlmssp module.

PR #48, opened 2018-05-28
smb2_set_password_from_file: fix for windows and uwp

afedchin:
getenv exists on Win32 but doesn't on UWP.


sahlberg:
Merged, thanks!

PR #60, opened 2018-07-08
Added symbolic link type

amosavian:
Interprets reparse points as symbolic links.

Considering a reparse point can't be passed to `smb2_unlink()`, `smb2_rename()`, etc methods, we should declare it as a separate type.

While reparse points can be different types than symlinks ([see here](https://en.wikipedia.org/wiki/NTFS_reparse_point#Features)) but I think the most common type are symbolic links. Regardless, the APIs for managing all of them are same.

We will implement functions regarding symbolic link later using ioctl interface.

### DIFF
--- a/include/smb2/libsmb2.h
+++ b/include/smb2/libsmb2.h
@@ -45,6 +45,7 @@ typedef void (*smb2_command_cb)(struct smb2_context *smb2, int status,
 /* Stat structure */
 #define SMB2_TYPE_FILE      0x00000000
 #define SMB2_TYPE_DIRECTORY 0x00000001
+#define SMB2_TYPE_SYMLINK   0x00000002
 struct smb2_stat_64 {
         uint32_t smb2_type;
         uint32_t smb2_nlink;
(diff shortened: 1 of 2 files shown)


amosavian:
I had more investigation and found out every system has a different method according to symbolic link.

My WD Mycloud device treats symbolic link as hard link, when I created using ssh.

macOS uses its own format for symbolic link when you create on remote server, but I’m not sure what about when when I use my macOS as server.

I did not investigate Windows yet.

amosavian:
Before your clarification, I’ve read SMB2 documentations and reparse points was promising as I thought this is a widely-implemented standard. Reparse points in SMB2 are either symlinks or mount-points.

From your comment I think reparse-points are limited to windows and probably macOS. Thus we must defer symlinks for now.

Thanks for your detailed comment.

haihongwang:
Currently Linux kernel SMB client does not support symlink operation:

# ln -s myfile linux_created_link.lnk
ln: failed to create symbolic link 'linux_created_link.lnk': Operation not supported

If a link is created on Windows (I believe it is reparse point), then it is shown as regular files:

-rwxr-xr-x 1 root root 746 Mar 21 16:41 windows_created_link.lnk

sahlberg:
Symbolic links works from linux if either

1, the server is samba and you force smb1 using   vers=1.0
This depends on samba offering unix capability, which is currently only
available for smb1.

2, you specify the mfsymlinks mount option.
This works with any server but is an ugly hack.

PR #64, opened 2018-08-05
determining literal width for constants

amosavian:
This PR helps to avoid unnecessary type casting when imported into type-safe languages like Swift. I'm not sure but it may micro-optimize performance.


amosavian:
@sahlberg  please review this pull request if possible

PR #80, opened 2019-02-12
Map more error status to EACCES

tguillem:
Otherwise, smb2 return an IO error.
This MR is useful for VLC that will ask the users for credential in case of EACCES or ECONNREFUSED errors.


sahlberg:
Merged, thanks

PR #83, opened 2019-04-28
Adding creation time to smb_stat

amosavian:
This PR adds file creation time to `smb2_stat` type. The naming is based on [linux's statx](http://man7.org/linux/man-pages/man2/statx.2.html) structure.


sahlberg:
Merged, thanks

PR #86, opened 2019-05-11
Support host:port syntax for specifying non-standard port numbers.

mwyoungpa:
This would allow an SMB share to be opened through a tunnel listening on another port, for instance.

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -75,8 +75,6 @@
 
 #define MAX_URL_SIZE 256
 
-#define CIFS_PORT 445
-
 static int
 smb2_get_credit_charge(struct smb2_context *smb2, struct smb2_pdu *pdu)
 {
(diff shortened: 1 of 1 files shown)


lib/socket.c
```diff
@@ -606,10 +605,18 @@ smb2_connect_async(struct smb2_context *smb2, const char *server,
...
                 }
                 *str = 0;
+                port = str + 1;
+        }
+
+        port = strchr(port, ':');
+        if (port != NULL) {
+                *port++ = 0;
+        } else {
+                port = "445";
```
amosavian: Why string?

sahlberg: A string and passing it to getaddrinfo() allows using service names and not just numbers.
I.e. /etc/services

sahlberg:
Merged, thanks.
Can you send a patch to document this too?
There are a few places where the url syntax is described. Those should be updated with the optional port number.

PR #88, opened 2019-05-15
Added error message on smb2_open failure

slevengood:
If smb2_open_async callback was given an unsuccessful status code, smb2_open would return NULL but no error message would be present. Error message is a copy of that from smb2_opendir.


sahlberg:
Merged, thanks.

PR #90, opened 2019-06-02
Add support for the ESP32 microcontroller

apuder:


sahlberg:
Merged, thanks!

sahlberg:
Merged, thanks.

Can you do a small update to README as well and add a couple of lines for
how/what to do to build it on that platform.
Just whatever someone else on ESP32 would need to get started.

PR #97, opened 2019-08-12
Android: fix getlogin_r() dep

tguillem:
* The getlogin_r() fallback was not working and not tested.
* It is also needed for Android before API 28

For Android, we could also fallback to getlogin() but I prefer returning an error here instead of using an unsafe function. Your call.

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -65,6 +65,14 @@
 #define getlogin_r(a,b) ENXIO
 #endif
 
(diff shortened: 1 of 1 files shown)


sahlberg:
Merged, thanks.

For android I think it is fine as is. I imagine that virtually every application that uses libsmb2 on android will set the username via the API anyway.

PR #104, opened 2019-10-17
errors: Add SMB2_STATUS_BAD_NETWORK_PATH

ePirat:
This error code is returned by macOS SMB implementation when not
specifying the share to which to connect.


sahlberg:
Merged, thanks

PR #105, opened 2019-10-17
Remove unused popt dependency for examples

ePirat:
The examples have a dependency on popt even though that library is
not used by any examples, making it unnecessary.


sahlberg:
merged. thanks

PR #111, opened 2019-10-17
Add windows build to CI

ePirat:
Currently without any of the optional dependencies (OpenSSL, GSSAPI), but better than nothing, I guess…

Fails currently, fix in #110


sahlberg:
Many thanks for setting these things up.

Don't worry about openssl and gssapi.
There are several other issues with those and I plan to eliminate these dependencies completely in the medium timeframe and replace them with a tiny but portable builtin implementation of kerberos for libsmb2.

PR #117, opened 2019-10-24
More errnos

tguillem:
This PR fixes invalid errno returned by smb2_connect_share_async() and add errnos to smb2_connect_async() return values.

FYI: smb2_connect_share_async() was returning -ENOMEM in case of I/O errors.

This will be used by VLC to know the reason of a connect failure and setup fallback SMBv1 fallback accordingly.


sahlberg:
Unmerged.

It doesn't build on error due to the second patch.
Please fix this and open a new PR.

tguillem:
Sure, no problem

PR #123, opened 2019-12-17
evaluate smb status before call to krb5 lib

amandeepgautam:
This leads to better error messages. For example, for incorrect pass we
have a message like following:
Session setup failed with (0xc000006d) STATUS_LOGON_FAILURE (-111)
with current change instead of
gss_init_sec_context: (Invalid token was supplied, Unknown error) (-1)

### DIFF
--- a/README
+++ b/README
@@ -133,8 +133,8 @@ by calling :
 
 SMB2/3 SIGNING
 ==============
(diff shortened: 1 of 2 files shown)


sahlberg:
Nice!
Thanks.

PR #128, opened 2020-01-08
add const qualifier for smb write related code

amandeepgautam:
This helps in making caller code free of casts. Also, at the caller side, in many cases, a guarantee of constantness is needed from the library to avoid copying of buffers.


amandeepgautam:
Thanks!

PR #130, opened 2020-01-31
extend error codes (from samba)

amandeepgautam:


sahlberg:
It still fails the linux build.  Can you have a look at what the issue is?

amandeepgautam:
@sahlberg it fails in dependencies stage. Not sure what is the reason

sahlberg:
Merged. Thanks.

I checked a random subset of the codes with MS-ERREF.PDF   which is now the canonical truth for these names and constants and they seem to match.

PR #131, opened 2020-02-01
Fix some mistake for build and using with VS2019

volehuy1998:
Nice library.

### DIFF
--- a/examples/smb2-share-enum.c
+++ b/examples/smb2-share-enum.c
@@ -1,3 +1,6 @@
+#pragma warning(disable: 4996)
+#pragma comment(lib, "Ws2_32")
+
 /* -*-  mode:c; tab-width:8; c-basic-offset:8; indent-tabs-mode:nil;  -*- */
 /*
    Copyright (C) 2016 by Ronnie Sahlberg <ronniesahlberg@gmail.com>
@@ -80,6 +83,20 @@ void se_cb(struct smb2_context *smb2, int status,
 
 int main(int argc, char *argv[])
 {
+	WORD wVersionRequested;
+	WSADATA wsaData;
+	int err;
+
+   	wVersionRequested = MAKEWORD(2, 2);
+
+   	err = WSAStartup(wVersionRequested, &wsaData);
+    	if (err != 0) {
+        	/* Tell the user that we could not find a usable */
+	        /* Winsock DLL.                                  */
+        	printf("WSAStartup failed with error: %d\n", err);
+	        return 1;
+    	}
+
         struct smb2_context *smb2;
         struct smb2_url *url;
 	struct pollfd pfd;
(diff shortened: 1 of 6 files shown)


volehuy1998:
I'm so interested in your lib

include/msvc/sys/socket.h
```diff
@@ -51,10 +51,10 @@ inline int readv(t_socket sock, struct iovec *iov, int nvecs)
   return -1;
 }
 
-inline int close(t_socket sock)
-{
-  return closesocket(sock);
-}
+//inline int close(t_socket sock)
```
sahlberg: Are you sure that this works on all versions of VS2019? 
Because it does build on earlier versions of VS without this patch.

A much safer change would be to have #ifdefs that just disables this for VS2019 or later.

lib/init.c
```diff
@@ -55,7 +55,7 @@
 #ifdef _MSC_VER
 #include <errno.h>
 #define getlogin_r(a,b) ENXIO
-#define srandom srand
+//#define srandom srand
```
sahlberg: Don't disable code by just commenting it out. Either delete it or #ifdef it out.

In this case, this works for earlier versions of VS so if you need to get rid of this define then you would need
to use an ifdef to ONLY remove it for VS2019 or later.

However, srand() has been part of the c standard since a very long time so if this fails to compile,  maybe you are missing some include you will need to bring in srand()

lib/init.c
```diff
@@ -270,7 +270,7 @@ void smb2_destroy_context(struct smb2_context *smb2)
         }
 
         if (smb2->fd != -1) {
-                close(smb2->fd);
+                fclose(smb2->fd);
```
sahlberg: These changes look wrong.   smb2->fd is a file descriptor, not a FILE * structure so this should not even compile.

examples/smb2-share-enum.c
```diff
@@ -80,6 +83,20 @@ void se_cb(struct smb2_context *smb2, int status,
 
 int main(int argc, char *argv[])
 {
+	WORD wVersionRequested;
```
sahlberg: This looks good but you will need to wrap this inside an ifdef so that it is only enabled when compiling on windows.

examples/smb2-share-enum.c
```diff
@@ -100,9 +117,9 @@ int main(int argc, char *argv[])
                         smb2_get_error(smb2));
                 exit(0);
         }
-        if (url->user) {
-                smb2_set_user(smb2, url->user);
-        }
+
+		smb2_set_user(smb2, "<usr>");
+		smb2_set_password(smb2, "<pwd>");
```
sahlberg: Remove this change or else the example will not work as a standalone program.

Wolf3s:
@sahlberg close this

PR #139, opened 2020-04-09
return appropriate error codes on failure

amandeepgautam:
with current change one can have same error handling logic for callbacks and
function return values. '-1' corresponds to permission denied which
makes it problematic.

### DIFF
--- a/lib/libsmb2.c
+++ b/lib/libsmb2.c
@@ -445,7 +445,7 @@ smb2_opendir_async(struct smb2_context *smb2, const char *path,
         dir = calloc(1, sizeof(struct smb2dir));
         if (dir == NULL) {
                 smb2_set_error(smb2, "Failed to allocate smb2dir.");
-                return -1;
+                return -EINVAL;
         }
         SMB2_LIST_ADD(&smb2->dirs, dir);
         dir->cb = cb;
@@ -465,7 +465,7 @@ smb2_opendir_async(struct smb2_context *smb2, const char *path,
         if (pdu == NULL) {
                 free_smb2dir(smb2, dir);
                 smb2_set_error(smb2, "Failed to create opendir command.");
-                return -1;
+                return -EINVAL;
         }
         smb2_queue_pdu(smb2, pdu);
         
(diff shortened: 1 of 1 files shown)


amandeepgautam:
For example, assume for lib call: smb2_opendir_async. The call would be like:
int rc = smb2_open_dir_async(smb2, .... , Cb1);

Cb1 will be defined as:
Cb1( int status, .....) {
  if (status) {
     ConvertErrorNumToOtherErrorFormat(status);
     return;
  }
}

Now, as of today, we cannot use ConvertErrorNumToOtherErrorFormat(rc) as '-1'  is returned at many places. This CL tries to return proper error code

amandeepgautam:
Also, I noticed that we do return EINVAL from some places in the lib. Ex: https://github.com/sahlberg/libsmb2/blob/0ddce21e13cd032adba26260f977b54d395f98c7/lib/socket.c#L660
So, I am hoping that this change will make return code consistent across the library.

lib/libsmb2.c
```diff
@@ -445,7 +445,7 @@ smb2_opendir_async(struct smb2_context *smb2, const char *path,
         dir = calloc(1, sizeof(struct smb2dir));
         if (dir == NULL) {
                 smb2_set_error(smb2, "Failed to allocate smb2dir.");
-                return -1;
+                return -EINVAL;
```
amandeepgautam, line 448: will change this to ENOMEM

lib/libsmb2.c
```diff
@@ -465,7 +465,7 @@ smb2_opendir_async(struct smb2_context *smb2, const char *path,
         if (pdu == NULL) {
                 free_smb2dir(smb2, dir);
                 smb2_set_error(smb2, "Failed to create opendir command.");
-                return -1;
+                return -EINVAL;
```
amandeepgautam, line 468: smb2_cmd_create_async can fail because of mem allocation or can fail because of failure during encoding, so I feel EINVAL is more appropriate Let me know your thoughts.

Issue #146, opened 2020-06-11
Timeout does not work on the connect() function because there are no PDUs in outqueue


timmeh87:
If I try to connect to a server that does not exist (make up a non-existent IP address) it seems that the connect() function never returns (infinite loop)

I believe this is why the timeout feature was recently added, to fix the infinite loop on sync functions, but I am unable to use it. It appears that the reason for this is that the "outqueue" is empty so that the "cb" can never actually be called

my example code:

```
	smb2_context * out = smb2_init_context();
	smb2_set_timeout(out, 1);
	struct smb2_url *url = smb2_parse_url(out, url_string.c_str());
	result = !(url == NULL);
	if (!result)
	{
		printf("Failed to parse url: %s\n", smb2_get_error(out));
	}
	smb2_set_password(out, password.c_str());
	smb2_set_security_mode(out, SMB2_NEGOTIATE_SIGNING_ENABLED);
	if (smb2_connect_share(out, url->server, url->share, url->user) != 0) {
		printf("smb2_connect_share failed. %s\n", smb2_get_error(out));
		result = false;
	}
```

When I pass this function a URL string for a server that does not exist, it never returns even though the timeout is set to "1"

I have debugged the code, and the reason for this is that the first time the code gets to "wait_for_reply()", at the line 

`smb2_timeout_pdus(smb2);`

The contents of smb2->outqueue is NULL

if you inspect the code inside that function, you will see that it is impossible to call the CB (and therefore exit the waiting loop) when this variable is NULL. 

Here is a picture of what I am seeing on the first pass through "wait_for_reply()" with the above example code

![image](https://user-images.githubusercontent.com/31858838/84431770-d95a4680-abf9-11ea-9a65-5ec9415ef3f8.png)

timmeh87:
Closing because my code was merged

PR #148, opened 2020-06-19
Improved timeout for sync interface when connection does not exist

timmeh87:
specifically, the "connect" function was not timing out if the remote did not ever reply. I added a rather "brute force" timeout to the waiting loop that will exit if there is no connection after the timeout. I believe that it should not affect anything else since there should normally **always** be a connection if we are waiting for a reply.


timmeh87:
Oh and it addresses #146

PR #151, opened 2020-08-08
only disconnect valid connections

amandeepgautam:
in such cases, we will now get response. previousy the call hung
indefinitely.


sahlberg:
merged. thanks

PR #163, opened 2020-10-09
fix build on APPLE

tguillem:
uuid_t is reserved by APPLE. It's a bad idea to add "_t" suffixes in
user program.

Apple need also time.h for time_t. Since C11, time_t is defined in
time.h, so not sure if all these ifdef are needed.

### DIFF
--- a/include/libsmb2-private.h
+++ b/include/libsmb2-private.h
@@ -23,7 +23,7 @@
 #include "config.h"
 #endif
 
-#if defined(PS2_EE_PLATFORM) || defined(PS3_PPU_PLATFORM) || defined(ESP_PLATFORM)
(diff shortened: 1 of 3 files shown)


include/libsmb2-private.h
```diff
@@ -23,7 +23,7 @@
 #include "config.h"
 #endif
 
-#if defined(PS2_EE_PLATFORM) || defined(PS3_PPU_PLATFORM) || defined(ESP_PLATFORM)
+#if defined(PS2_EE_PLATFORM) || defined(PS3_PPU_PLATFORM) || defined(ESP_PLATFORM) | defined(__APPLE__)
```
ePirat, line 26: Should be `||`

ePirat (changes requested):

PR #164, opened 2020-10-14
handle STOPPED_ON_SYMLINK error code correctly

amandeepgautam:


amandeepgautam:
-- if not done, we do not get the callback in our application code.
-- Attaching the response of a create request:
```
SMB2 (Server Message Block Protocol version 2)
    SMB2 Header
        ProtocolId: 0xfe534d42
        Header Length: 64
        Credit Charge: 1
        NT Status: STATUS_STOPPED_ON_SYMLINK (0x8000002d)
        Command: Create (5)
        Credits granted: 42
        Flags: 0x00000001, Response
            .... .... .... .... .... .... .... ...1 = Response: This is a RESPONSE
            .... .... .... .... .... .... .... ..0. = Async command: This is a SYNC command
            .... .... .... .... .... .... .... .0.. = Chained: This pdu is NOT a chained command
            .... .... .... .... .... .... .... 0... = Signing: This pdu is NOT signed
            .... .... .... .... .... .... .000 .... = Priority: This pdu does NOT contain a PRIORITY
            ...0 .... .... .... .... .... .... .... = DFS operation: This is a normal operation
            ..0. .... .... .... .... .... .... .... = Replay operation: This is NOT a replay operation
        Chain Offset: 0x00000000
        Message ID: 4
        Process Id: 0x0000feff
        Tree Id: 0x00000001  \\10.2.203.12\abcd
            [Tree: \\10.2.203.12\abcd]
            [Share Type: Physical disk (0x01)]
            [Connected in Frame: 38]
        Session Id: 0x020c037400001fd3
            [Account: Administrator Acct:Administrator]
            [Domain: QA01 Domain:QA01]
            [Host: U18-100302-01 Host:U18-100302-01]
            [Authenticated in Frame: 35]
        Signature: 00000000000000000000000000000000
        [Response to: 40]
        [Time from request: 0.000286000 seconds]
    Create Response (0x05)
        StructureSize: 0x0009
            0000 0000 0000 100. = Fixed Part Length: 4
            .... .... .... ...1 = Dynamic Part: True
        Error Context Count: 0
        Reserved: 0x00
        Byte Count: 64
        Error Data: 3c00000053594d4c0c0000a0300000001200120000001200…
            Symbolic Link Error Response
                SymLink Length: 60
                SymLink Error Tag: 0x4c4d5953
                Reparse Tag: REPARSE_TAG_SYMLINK (0xa000000c)
                Reparse Data Length: 48
                Unparsed Path Length: 0
                Flags: 1
                Substitute Name: file1.txt
                    Blob Offset: 0x00000012
                    Blob Length: 18
                Print Name: file1.txt
                    Blob Offset: 0x00000000
                    Blob Length: 18
````
-- Samba handles this error code.

File: lib/pdu.c [hunk 1]
```diff
@@ -397,15 +397,23 @@ smb2_find_pdu(struct smb2_context *smb2,
...
+        if ((smb2->hdr.status & SMB2_STATUS_ERROR_SEVERITY_MASK) ==
             SMB2_STATUS_SEVERITY_ERROR) {
                 switch (smb2->hdr.status) {
                 case SMB2_STATUS_MORE_PROCESSING_REQUIRED:
                         return 0;
                 default:
                         return 1;
                 }
-        }
+        } else if ((smb2->hdr.status & SMB2_STATUS_WARNING_SEVERITY_MASK) ==
```
amandeepgautam, line 408: -- not sure if this is the best way to handle this. Please suggest an alternative.
-- I am not familiar with samba code base so not sure how it is handled there.
-- also, there are many other warning codes from SMB specs which are not handled here. For now, I have just taken the conservative approach: kept the existing behaviour for all warning codes except the one I faced a problem with.

lib/errors.c
```diff
@@ -1117,6 +1119,7 @@ int nterror_to_errno(uint32_t status) {
         case SMB2_STATUS_INVALID_PARAMETER:
         case SMB2_STATUS_NOT_SUPPORTED:
         case SMB2_STATUS_NOT_A_REPARSE_POINT:
+        case SMB2_STATUS_STOPPED_ON_SYMLINK:
```
amandeepgautam, line 1122: Not sure if this is the best posix error code mapping. Suggestions.

amandeepgautam:
@sahlberg  Please review. added a couple of questions for you to answer. thanks!

include/smb2/smb2-errors.h
```diff
@@ -19,11 +19,12 @@
 /*
  * NTSTATUS fields
  */
-#define SMB2_STATUS_SEVERITY_MASK    0xc0000000
-#define SMB2_STATUS_SEVERITY_SUCCESS 0x00000000
-#define SMB2_STATUS_SEVERITY_INFO    0x40000000
-#define SMB2_STATUS_SEVERITY_WARNING 0x80000000
-#define SMB2_STATUS_SEVERITY_ERROR   0xc0000000
+#define SMB2_STATUS_ERROR_SEVERITY_MASK    0xc0000000
+#define SMB2_STATUS_WARNING_SEVERITY_MASK  0x80000000
```
sahlberg, line 23: No, leave it as it was before.
The severity bitmask is the top two bits and it encods 4 different values,  success/info/warning and error.

So the way to test if something is an error is thus first to mask out the two severity bits and then check what these two bits are,
for example

switch (status & SMB2_STATUS_SEVERITY_MASK) {
case SMB2_STATUS_SEVERITY_SUCCESS:  printf("success!\n"); break;
case SMB2_STATUS_SEVERITY_ERROR: printf("was error\n"); break;
...

lib/pdu.c
```diff
@@ -397,15 +397,23 @@ smb2_find_pdu(struct smb2_context *smb2,
 static int
 smb2_is_error_response(struct smb2_context *smb2,
                        struct smb2_pdu *pdu) {
-        if ((smb2->hdr.status & SMB2_STATUS_SEVERITY_MASK) ==
+        if ((smb2->hdr.status & SMB2_STATUS_ERROR_SEVERITY_MASK) ==
```
sahlberg, line 400: No, remove this change.

lib/pdu.c
sahlberg: This should just be } else if ((smb2->hdr.status & SMB2_STATUS_SEVERITY_MASK) ==

Here is why this would be wrong.
Imagine you have a status 0xc0000016  which is an error, not a warning.

IF you only check this mask using a single bit for the bitmask, i.e. SMB2_STATUS_WARNING_SEVERITY_MASK==0x80000000

Then this code becomesITY_MASK
if ((status & SMB2_STATUS_WARNING_SEVERITY_MASK) == SMB2_STATUS_SEVERITY_WARNING)
which would be

But status/0xc0000016 masked with SMB2_STATUS_WARNING_SEVERITY_MASK/0x8000000
which is equal to SMB2_STATUS_SEVERITY_WARNING

I.e.   0xc0000016 is an error, not a warning, but this if statement would still be true for errors.

sahlberg, line 408: Aside from the changes to the MASKS, remove these changes, otherwise the patch looks fine.
Please update it and I will merge it.

Thanks!

sahlberg (changes requested):

include/smb2/smb2-errors.h
amandeepgautam: done.

lib/pdu.c
```diff
@@ -405,6 +405,14 @@ smb2_is_error_response(struct smb2_context *smb2,
                 default:
                         return 1;
                 }
+        } else if ((smb2->hdr.status & SMB2_STATUS_SEVERITY_MASK) ==
+                 SMB2_STATUS_SEVERITY_WARNING) {
+                switch(smb2->hdr.status) {
+                case SMB2_STATUS_STOPPED_ON_SYMLINK:
```
amandeepgautam, line 411: Do you think there are other warnings which need to be handled?

amandeepgautam:
@sahlberg addresses your comments. ptal.

sahlberg:
Merged. Thanks

PR #170, opened 2020-10-29
Don't send a sign error if it is not required

tguillem:
Tested on a SMB3 server (samba 4.12.5-Debian) with the following
smb.conf:

[share]
...
public = yes


sahlberg:
Merged. Thanks.

PR #173, opened 2020-11-12
RFC 8305: Happy eyeballs

tguillem:
Hello,

Lot of VLC for android and iOS users have trouble connecting to some servers/NAS. From my investigation, some NAS refuse IPv4, some other refuse IPv6. Some NAS have firewall dropping a port for one address family, resulting on a very long timeout when try to connect to it.

That is what the Happy eyeballs RFC is trying to solve, cf. https://tools.ietf.org/html/rfc8305

This PR is inspired from Martin work on FFmpeg: http://git.videolan.org/?p=ffmpeg.git;a=commitdiff;h=9b4c3f5aadf54ffd2a6e15746b1fd736379883c4

PS: Videolan will release new BETAs for Android and iOS using this new branch. I suggest waiting for some feedbacks before merging it.

### DIFF
--- a/include/libsmb2-private.h
+++ b/include/libsmb2-private.h
@@ -127,7 +127,6 @@ struct smb2_context {
         t_socket *connecting_fds;
         size_t connecting_fds_count;
 
-        int is_connected;
         int timeout;
 
         enum smb2_sec sec;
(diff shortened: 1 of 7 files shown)


amandeepgautam:
@tguillem This looks promising. Any idea by when will this be merged?

Issue #185, opened 2021-01-13
smb2_read_data fails on MacOS (10.15.7) shared files


toto263:
smb2_read_data fails on macOS shares.
After some testing, I found that the error occurs at the followings.

I don't know why, but signature verification fails on mac, and it return "Wrong signature in received" error.
Remove the verification (memcmp), then works ok.

```
socket.c - smb2_read_data
    ...     
        /* We don't yet have the signing key until later, once session
         * setup has completed, so we can not yet verify the signature
         * of the final leg of session setup.
         */
        if (smb2->sign &&
            (smb2->hdr.flags & SMB2_FLAGS_SIGNED) &&
            (smb2->hdr.command != SMB2_SESSION_SETUP) ) {
                uint8_t signature[16];
                memcpy(&signature[0], &smb2->in.iov[1].buf[48], 16);
                if (smb2_calc_signature(smb2, &smb2->in.iov[1].buf[48],
                                        &smb2->in.iov[1],
                                        smb2->in.niov - 1) < 0) {
                        return -1;
                }

//--> works if commented out the followings
                if (memcmp(&signature[0], &smb2->in.iov[1].buf[48], 16)) {
                        smb2_set_error(smb2, "Wrong signature in received "
                                       "PDU");
                        return -1;
                }
        }

...
```

sahlberg:
So when signing is on, the server accepts the signatures that libsmb2 sends to it  but libsmb2 rejects the signatures that mac sends to us?

Can you collect a network trace of when this fails as well as a network trace of when it works when you comment out the verification?
Can you also try forcing an older dialect in libsmb2 (you can set version via url argument)
for example try one of the 2.x dialects as well as one of the 3.x dialects (but not 3.1.1 which is where we have negotiation of signing algorithms)

toto263:
I tried all dialects using "smb2_set_version(context, SMB2_VERSION_xxxx)", from SMB2_VERSION_0202 to SMB2_VERSION_0302, but failed.

I found that another signature verification in session_setup_cb() of libsmb2.c, and this works ok.
```
libsmb2.c - session_setup_cb()
    ...
                if (smb2->hdr.flags & SMB2_FLAGS_SIGNED) {
                        uint8_t signature[16];

                        memcpy(&signature[0], &smb2->in.iov[1].buf[48], 16);
                        if (smb2_calc_signature(smb2, &smb2->in.iov[1].buf[48],
                                                &smb2->in.iov[1],
[...]

Issue #183, opened 2021-01-05
Parallel connections to shares of the same server fail.


s-devyatkin:
Attempts to concurrently connect to share (or different shares) of the same server by smb2_connect_share or smb2_connect_share_async (with different smb2_context, of course) fail with error "Wrong signature in received PDU". The error is generated in libsmb2.c. Using mutex on smb2_connect_share solves this problem, but not for for async version. Maybe I'm wrong, but it looks like a bug. Detected on Debian 9.

One more question. Is there a way to use a single connection to the server for all its shares?

delins:
Do you still have this problem? I have a similar use case, running on Ubuntu 18.04, but I haven't had any issues connecting to multiple shares asynchronously (both on the same machine and different ones). 

Could you attach an mwe that reproduces the problem? It could also be useful to explicitly set all of `smb2_set_security_mode`, `smb2_set_seal`, `smb2_set_sign` and `smb2_set_authentication` in the mwe so that it's clearer what connection parameters may cause it.

s-devyatkin:
```
#include <stdint.h>
#include <string.h>
#include <iostream>
#include <future>
#include "smb2.h"
#include "libsmb2.h"

void task(const char* server, const char* share, const char* path, const char* user, const char* password)
{
    struct smb2_context *smb2;
    struct smb2_stat_64 st;

    smb2 = smb2_init_context();
    if (smb2 == NULL)
    {
        std::cerr << "Failed to init context" << std::endl;
        return;
    }

    smb2_set_user(smb2, user);
    smb2_set_password(smb2, password);
    smb2_set_sign(smb2, 1);
    smb2_set_seal(smb2, 1);
    smb2_set_security_mode(smb2, SMB2_NEGOTIATE_SIGNING_ENABLED);

    if (smb2_connect_share(smb2, server, share, 0) != 0)
    {
        std::cerr << "smb2_connect_share failed: " << smb2_get_error(smb2) << std::endl;
        return;
    }

    if (smb2_stat(smb2, path, &st) < 0)
    {
        std::cerr << "smb2_stat failed: " << smb2_get_error(smb2) << std::endl;
        return;
    }

    std::cerr << "success" << std::endl;

    smb2_disconnect_share(smb2);
    smb2_destroy_context(smb2);
}

int main(int argc, char *argv[])
{
    auto t1 = std::async(std::launch::async, &task, "server", "share", "file", "user", "password");
[...]

Issue #176, opened 2020-11-17
Wrong signature in received PDU


tguillem:
When connecting to a samba Version 4.12.5-Debian with the following configuration:

```
min protocol = SMB2
max protocol = SMB2
server signing = mandatory
```

smb2_service() fails with the following error: "Wrong signature in received PDU".

The error come from smb2_read_data() in lib/socket.c.

Note that the PDU signature check is OK from session_setup_cb() in lib/libsmb2.c

Removing the PDU signature check from lib/socket.c (and letting it in lib/libsmb2.c) is fixing my issue.

What is the proper solution?

delins:
@tguillem A similar issue was fixed in #183. Could you check if the latest commit fixes this issue as well?

tguillem:
@delins I confirm that the lasted commit fixes this issue. Thanks a lot !

PR #192, opened 2021-01-27
Fix signature validation error

delins:
Should fix the signature validation error discussed in #183 and possibly #185 and #176


sahlberg:
Merged, thanks

PR #193, opened 2021-01-27
Make globals local in aes code

delins:
Should be part of the fix needed to address the multi threading related problem in #183 

My apologies for the many unrelated lines that were modified: they had a trailing whitespace and my editor likes to remove them.

I imagine that previously `AES128_CBC_encrypt_buffer` and `AES128_CBC_decrypt_buffer` were to be called with an actual iv the first time, and with 0 any time after that until encryption was complete. The code in aes.c kept the `iv` state in a global variable.

The global variable is gone, so the caller is responsible for keeping the iv buffer alive. I think it should be added to `smb2_context` but since the CBC code isn't used anywhere I've not yet added it as an (`ifdef`fed) member of `smb2_context`. You may want to modify the pr and add a comment to aes.h about it.

### DIFF
--- a/lib/aes.c
+++ b/lib/aes.c
@@ -16,10 +16,10 @@ ECB-AES128
   key:
     2b7e151628aed2a6abf7158809cf4f3c
   resulting cipher
-    3ad77bb40d7a3660a89ecaf32466ef97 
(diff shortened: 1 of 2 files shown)


sahlberg:
Merged. Thanks

PR #197, opened 2021-04-14
Using random value for client_guid instead of process id

wi7a1ian:
This fixed an issue with Microsoft Windows 10 SMB server returning `STATUS_USER_SESSION_DELETED` when opening multiple connections from within a service hosted in docker container (or multiple containers I would rather say). It was happening only when SMB protocol 3.1.1 was used. Negotiating 3.0.2 or lower version did also help, but that isn't good candidate for long term solution.


sahlberg:
Merged. Thanks.

Issue #190, opened 2021-01-18
Add opaque pointer member to smb2_context


delins:
A suggestion to add an opaque pointer to `smb2_context`, which users of the library can set similarly to `smb2_fd_event_callbacks` (eg by calling `smb2_set_opaque(smb2, myOpaquePointer)`. This can then be read from the `smb2_context` again in the callbacks using something like `smb2_get_opaque(smb2)`.

This makes it possible to bring state into the callbacks, which means we don't have to work with globals like `is_finished`, `cfd` and `cevents` in `smb2-ls-async.c`. The clear benefit of this is that we can reuse the callback function for multiple `smb2_context`s.

It is unfortunate though that `smb2_context` is private to the library, which means that a call like `smb2_get_opaque()` cannot be inlined. The extra call could be avoided if the caller of the callbacks would add the opaque pointer in its call, but that's not possible or the callback signatures would need to change. So I'm not sure whether the convenience would outweigh the overhead. What are your thoughts?

delins:
I've implemented the version with the modified function signatures. It's so little work that I'd be fine with locally applying this change after each pull.

delins:
Nevermind, the `cb_data` parameter does just that and more. I'd seen it before but it's not documented, and looking at some of the examples it seemed like it needed to be passed specific data. It's also called "private_data" in the examples, which suggests that it may be private to the library (see also `private` in OOP languages).

I guess it's all completely obvious on first sight to people who are used to asynchronous programming, but it may still be a good idea to document it for those diving in for the first time :)

delins:
`smb2_set_opaque` and `smb2_get_opaque` merged in #201, thanks!

PR #201, opened 2021-04-24
add opaque context member to smb2_context

delins:
Implements #190. 

libsmb2 already allows context to be passed to async functions, for instance a directory name in the callback for an async directory listing. However, if you want to use the library in an OOP way, you always want some state object to be reachable from within a callback. In that case you'd need to wrap both in another struct, which creates overhead and is just cumbersome.

This commit basically allows you to set and get this state object anywhere you want, freeing the `private_data` for the thing that is specific to that async call.

It can also be called something other than `opaque`.

### DIFF
--- a/include/libsmb2-private.h
+++ b/include/libsmb2-private.h
@@ -148,6 +148,8 @@ struct smb2_context {
         const char *workstation;
         char client_challenge[8];
 
+        void *opaque;
+
         smb2_command_cb connect_cb;
         void *connect_data;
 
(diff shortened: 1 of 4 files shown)


sahlberg (changes requested):
Need to add smb2_set/get_opaque to libsmb2.syms too

sahlberg (approved):

sahlberg:
merged, thanks

PR #202, opened 2021-06-11
Fix shared folder with ; or @ issue

### LINKED ISSUE #84: smb2_parse_url parse has bug when no domain.
Smb2_parse_url parse has a bug,  when there is no domain in url bug path has ';' or '@', 

"smb2_parse_url"  should first extract the smb:// to the path between smb://domain;user:ip/path, and then analyze the domain and user pwd and ip.

struct smb2_url *smb2_parse_url(struct smb2_context *smb2, const char *url)
{
        struct smb2_url *u;
        char *ptr, *tmp, str[MAX_URL_SIZE];
        char *args;

        if (strncmp(url, "smb://", 6)) {
                smb2_set_error(smb2, "URL does not start with 'smb://'");
                return NULL;
        }
        if (strlen(url + 6) >= MAX_URL_SIZE) {
                smb2_set_error(smb2, "URL is too long");
                return NULL;
        }
	strncpy(str, url + 6, MAX_URL_SIZE);

        args = strchr(str, '?');
        if (args) {
                *(args++) = '\0';
                if (smb2_parse_args(smb2, args) != 0) {
                        return NULL;
                }
        }

        u = malloc(sizeof(struct smb2_url));
        if (u == NULL) {
                smb2_set_error(smb2, "Failed to allocate smb2_url");
                return NULL;
        }
        memset(u, 0, sizeof(struct smb2_url));

        ptr = str;

        /* domain */
        if ((tmp = strchr(ptr, ';')) != NULL) {  //this will be bug when path is smb://192.168.1.200/path/1;abc.txt
                *(tmp++) = '\0';
                u->domain = strdup(ptr);
                ptr = tmp;
        }
        /* user */
        if ((tmp = strchr(ptr, '@')) != NULL) { // the same bug like above.
                *(tmp++) = '\0';
                u->user = strdup(ptr);
                ptr = tmp;
	}
        /* server */
        if ((tmp = strchr(ptr, '/')) != NULL) {
                *(tmp++) = '\0';
                u->server = strdup(ptr);
                ptr = tmp;
        }

        /* Do we just have a share or do we have both a share and an object */
        tmp = strchr(ptr, '/');
        
        /* We only have a share */
        if (tmp == NULL) {
                u->share = strdup(ptr);
                return u;
        }

        /* we have both share and object path */
        *(tmp++) = '\0';
        u->share = strdup(ptr);
        u->path = strdup(tmp);

        return u;
}

NicolasLQ:
smb2_parse_url function needs an URL with the format :
smb://[<domain;][<username>@]<server>/<share>/<path>

When optional args are not provided, the parser will still search them
and maybe find them in the <share>/<path>. If the delimiters ; or @
are found in it, this will pose a problem to analyze the parsing.

Fixes #84

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -210,14 +210,21 @@ struct smb2_url *smb2_parse_url(struct smb2_context *smb2, const char *url)
 
         ptr = str;
 
+        char *shared_folder = strchr(ptr, '/');
+        if (!shared_folder) {
+                smb2_set_error(smb2, "Wrong URL format");
+                return NULL;
+        }
+        int len_shared_folder = strlen(shared_folder);
+
         /* domain */
-        if ((tmp = strchr(ptr, ';')) != NULL) {
+        if ((tmp = strchr(ptr, ';')) != NULL && strlen(tmp) > len_shared_folder) {
                 *(tmp++) = '\0';
                 u->domain = strdup(ptr);
                 ptr = tmp;
         }
         /* user */
-        if ((tmp = strchr(ptr, '@')) != NULL) {
+        if ((tmp = strchr(ptr, '@')) != NULL && strlen(tmp) > len_shared_folder) {
                 *(tmp++) = '\0';
                 u->user = strdup(ptr);
                 ptr = tmp;


sahlberg:
Merged, thanks

PR #211, opened 2021-09-06
Check for time.h and include as appropriate

SeanMcG:
musl libc defines time_t in this header


Wolf3s:
@SeanMcG You should add these for the makefiles

SeanMcG:
> @SeanMcG You should add these for the makefiles

I'm not sure I understand, can you elaborate?

sahlberg:
This looks good.

sahlberg:
merged, thanks

Issue #204, opened 2021-06-24
Windows cmake DLL build fails to produce import library - "dlldir" is not being used on windows


timmeh87:
On Windows, when downloading the latest code and running CMAKE on it, a visual project is produced and the build succeeds. However there is no ".lib" file produced because none of the functions are explicitly exported. This is 'normal' expected behavior on windows and means that libsmb2 can only be used as a static library on windows.

Cmake provides a workaround called "WINDOWS_EXPORT_ALL_SYMBOLS" which could be used to fix this problem, although some people think it is "not the best way to do things"

I only bring this up because I have recently ported libsmb2 into the vcpkg project 
https://github.com/microsoft/vcpkg/issues/18454

While the port was accepted, it was not accepted to pass "WINDOWS_EXPORT_ALL_SYMBOLS" from the vcpkg side, as this is against their own policy. I was really hoping to have a DLL version of libsmb2 available on vcpkg, which means that libsmb2 itself has to support windows DLL builds somehow. 

I would be willing to make a PR to address this but I would like to know the best course of action @sahlberg

sahlberg:
Don't export all symbols. We have a list of the symbols that should be exported and is is used in the autotool bulds :
-export-symbols ${srcdir}/libsmb2.syms

I don't know much about building on windows, but it should be possible to tell the cmake windows build to :
only export symbols from libsmb2.syms?

As said, I don't know much about windows builds but we do have a list of symbols to be exported that is used everywhere else except windows + cmake build?   Patches welcome.

timmeh87:
I think you are right, I will look into this

shalvenlaw:
AFAIK, there are two ways to export symbols from dll in Visual Studio:
1. using **__declspec(dllexport)** keyword
2. using [*.def file](https://docs.microsoft.com/en-us/cpp/build/reference/module-definition-dot-def-files?view=msvc-170)

Since *.def file in Windows is something just like *.syms file in Linux, we can change **libsmb2.syms** to **libsmb2.def** easily. This is my solution:
1. add a *.def file to the solution in Visual Studio 2017, named libsmb2.def
2. copy the content form libsmb2.syms to libsmb2.def, do some modifications according to the [rules](https://docs.microsoft.com/en-us/cpp/build/reference/rules-for-module-definition-statements?view=msvc-170) and [example](https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files?view=msvc-170)
[...]

PR #225, opened 2022-04-29
Add support for syms file when using msvc

shalvenlaw:
PR for [issue#204](https://github.com/sahlberg/libsmb2/issues/204): I added some lines in CMakeLists.txt, so that when using msvc to build shared library, CMake will read the libsmb2.syms to generate a libsmb2.def and add it to the VS solution. 

This will enable the generation of .lib file(that is, the exported symbols from dll) when building the VS solution.


sahlberg:
Merged, thanks

PR #233, opened 2022-08-09
fix mismatched bound build error

tguillem:
With gcc (Debian 11.3.0-4) 11.3.0:

../../lib/sha224-256.c:341:48: error: argument 2 of type ‘uint8_t[]’ {aka ‘unsigned char[]’} with mismatched bound [-Werror=array-parameter=]
  341 | SHA256Result (SHA256Context * context, uint8_t Message_Digest[])
      |                                        ~~~~~~~~^~~~~~~~~~~~~~~~

### DIFF
--- a/lib/sha224-256.c
+++ b/lib/sha224-256.c
@@ -338,7 +338,7 @@ SHA256FinalBits (SHA256Context * context,
  *   sha Error Code.
  */
 int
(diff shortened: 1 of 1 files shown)


sahlberg:
Lol, I also just pushed the same fix.
Thanks for the fix.  Merged.

PR #235, opened 2022-10-27
build/cmake: only include OpenSSL include directory if it is found

uyjulian:
Fixes CMake build when OpenSSL is not found

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,8 +33,11 @@ include(cmake/ConfigureChecks.cmake)
 
 include_directories(${CMAKE_CURRENT_BINARY_DIR}
                     include
-                    include/smb2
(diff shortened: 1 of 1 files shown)


sahlberg:
Libsmb2 does not use or depend on openssl.
Openssl is not even be compatible with lgplv2 so we should and can not use it.

llakssz:
Please reopen, I could not build on Windows as `OPENSSL_INCLUDE_DIR` was not set.
Either installing openssl and setting this variable, or using this fix submitted fixes the issue.
Probably it has not been encountered often as most people build libsmb2 for other platforms.
Thanks

delins:
I think this should be fixed. Libsmb2 may not use openssl, but it does `findpackage` it in CMakeLists.txt.

PR #239, opened 2022-11-29
add support for optional password in url

tmaillart:


sahlberg:
I am not a really big fan of this for several reasons.
Passwords should never really be visible on the commandline because that means they will leak via vectors like 'ps aux' or a lot of times applications that invoke commands that put the commandlines for childprocesses in their log files.

Any reason why you can not use the NTLM_USER_FILE mechanic?

tmaillart:
I understand your point, and I agree, conceptually it is wrong.

However in my use case,
1. process **A** presents the content of the smb share
2. the user selects a file
3. **A** passes a libsmb2 url to the player (process **B**)

And nobody should be able to execute arbitrary code on this machine.

sahlberg:
> I understand your point, and I agree, conceptually it is wrong.
> 
> However in my use case,
> 
>     1. process **A** presents the content of the smb share
> 
>     2. the user selects a file
> 
>     3. **A** passes a libsmb2 url to the player (process **B**)
> 
> 
> And nobody should be able to execute arbitrary code on this machine.

Yeah, but no. If this goes into the master branch it will affect all platforms, so a no on the pr.
Why can't you just use the NTLM_USER_FILE mechanics?  
Is getenv() not available on your platform?  Should be easy to write a replacement.  Or just store the password somewhere else and use smb2_set_password() on the context before connecting?

Issue #242, opened 2023-01-04
Broken readv/writev implementation


averne:
Hi, 

While porting libsmb2 to the Nintendo Switch, I came accross an issue caused by the `readv`/`writev` fallback implementations found in [compat.c](https://github.com/sahlberg/libsmb2/blob/master/lib/compat.c). I'm reporting this as a heads up, if anyone ever encounters the same situation.

The problem is that these implementations will call the relevant functions on a loop in the case of a partial read/write instead of returning the partial amount of data read/written, as would the normal read/write functions do. According to the manpage, "The `readv()` function shall be equivalent to `read()` [...]". In particular, listing large directories made the library hang (calling `poll` indefinitely).

Fixing this behavior allowed the library to progress further, but with very slow transfer speeds compared to what I achieve on other systems. I didn't try to pin down the problem to `readv` or `writev`, but ended up ripping out the [glibc implementations](https://github.com/bminor/glibc/blob/master/sysdeps/posix/readv.c) which made everything work as expected.

sedgers:
Hi averne,

Thanks hugely for this observation.  I am using libsmb2 with an ESP32 NodeMCU board running ESP-IDF SDK v4.4.3 to read files from a Western Digital NAS drive over WiFi.  Small files work, but when file size exceeds the read buffer and the read is chunked, I see the same issue with calling readv then hanging on poll() waiting for an event that never arrives.   Until now I've spent a few days playing with this without success, other than noting that by slowing things down by turning on debug I can improve things intermittently - guess this affects the dynamics of filling the readv buffer.

ESP32 ESP-IDF does not provide readv so the compat.c implementation gets used as per your Nintendo.  I changed the implementation to reflect the flow of control in the glibc version of readv as you suggest, and things spring into life.  I haven't tested this exhaustively yet but I am pretty sure you have nailed the cause here.  Hopefully Ronnie will be interested in your bug fix, as being able to read NFS files on ESP32 is wonderful.

sahlberg:
Can you send a pull request?

averne:
See https://github.com/sahlberg/libsmb2/pull/243/files

averne:
Closing after the merge of https://github.com/sahlberg/libsmb2/pull/243

PR #243, opened 2023-01-09
Fix fallback readv/writev implementations

averne:
Reasoning described in https://github.com/sahlberg/libsmb2/issues/242.

The code is adapted from glibc. I removed the alloca optimization as I was concerned it wouldn't build on the various platforms libsmb2 supports, let me know if I should add it back. This was only tested on Linux.

Since this implementation unconditionally allocates and copies, an easy  optimization would be to directly call read if the number of vectors is 1. Let me know as well.

### DIFF
--- a/lib/compat.c
+++ b/lib/compat.c
@@ -205,52 +205,88 @@ void smb2_freeaddrinfo(struct addrinfo *res)
 #endif /* PS3_PPU_PLATFORM */
 
 #ifdef NEED_WRITEV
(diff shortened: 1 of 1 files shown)


sahlberg:
Merged. Thanks!
Minor nit is that many older compilers might not support mixing declaration and code, but if that happens it will be trivial to fix it up and just move the declarations of char *buffer to the beginnning of the function.

PR #247, opened 2023-01-17
give access to nterror

tmaillart:
Most of the time, the real **nterror** is much more explicit than a mapping to errors defined in **errno.h**.

Here is a patch to be able to get this error, if it is not related to the protocol you get 0 (== success).


sahlberg:
Merged, thanks

PR #248, opened 2023-01-17
Fixed access mode checking to also work on platforms where O_RDWR is defined as (O_RDONLY|O_WRONLY).

salass00:


sahlberg:
O_ACCMODE is not universally available. Can you update the patch to handle this?

salass00:
Done.

sahlberg:
Merged, thanks

PR #249, opened 2023-01-18
Fix ntlm_convert_password_hash on big endian

BSzili:
Using NTLM hashes didn't work on big endian systems, because utf8_to_utf16 always converts to little endian. A byteswap is required before parsing the hash string on these systems.


sahlberg:
Merged, thanks

PR #258, opened 2023-03-28
Add ESP-IDF 5.x support

oliverschmidt:
https://docs.espressif.com/projects/esp-idf/en/v5.0.1/esp32/migration-guides/release-5.x/system.html#esp-system says:

The header files esp_random.h, esp_mac.h, and esp_chip_info.h, which were all previously indirectly included via the header file esp_system.h, must now be included directly. These indirect inclusions from esp_system.h have been removed.


sahlberg:
Merged, thanks!

PR #259, opened 2023-04-08
Update CMakeLists.txt to remove OpenSSL references

llakssz:
Basically a copy from https://github.com/sahlberg/libsmb2/pull/235
Only add openssl include directory if found.
This fix allows me to build on Windows, without having OpenSSL installed.

You say "Libsmb2 does not use or depend on openssl"-  then I guess we should remove all references to it entirely, but not sure if there's some other files that need to be altered too.


llakssz:
Should be good now!

sahlberg:
merged, thanks

PR #261, opened 2023-05-22
Fix bounds checks to where we read fields from the NTLM buffer

Rob-McKay:
Fix the bounds checks to where we read fields from the NTLM buffer when parsing the challenge packet from the server.


sahlberg:
Merged, thaks

PR #265, opened 2023-06-10
Connection data for context is automatically freed when context is destroyed

healytpk:
In my network analysis program, **_Dynamo_**, I have a thread dedicated to sending packets out on the network. This thread runs through a few coroutines in series, and one of the coroutines sends a '**_Samba Enum List_**' query to every host on the network. The asynchronous API in **_libsmb2_** is perfect for this.
However I have coded a timeout into **_Dynamo_**. If the list of shares hasn't been received within 5 seconds, the connection to the Samba server on TCP port 445 is instantly closed by calling '**_smb2_destroy_context_**'. This can cause a memory leak of the '**_connection_data_**' structure that gets allocated inside '**_smb2_connect_share_async_**'.
The purpose of this patch is to allow instantaneous closure of the TCP socket without a memory leak.

I have minimal experience with **_libsmb2_**, I only started using it 2 days ago -- so maybe I'm just using the library incorrectly and causing the memory leak myself.

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -362,6 +362,10 @@ void smb2_destroy_context(struct smb2_context *smb2)
         free(discard_const(smb2->workstation));
         free(smb2->enc);
 
+        if (smb2->connect_data) {
+            free_c_data(smb2, smb2->connect_data);  /* sets smb2->connect_data to NULL */
+        }
+
         free(smb2);
 }
 
--- a/lib/libsmb2.c
+++ b/lib/libsmb2.c
@@ -506,7 +506,7 @@ smb2_opendir_async(struct smb2_context *smb2, const char *path,
         return 0;
 }
 
-static void
+extern void
 free_c_data(struct smb2_context *smb2, struct connect_data *c_data)
 {
         if (c_data->auth_data) {
@@ -526,6 +526,10 @@ free_c_data(struct smb2_context *smb2, struct connect_data *c_data)
         free(discard_const(c_data->share));
         free(discard_const(c_data->user));
         free(c_data);
+
+        if (smb2->connect_data == c_data) {
+            smb2->connect_data = NULL;  /* to prevent double-free in smb2_destroy_context */
+        }
 }
 
 
--- a/lib/libsmb2.c
+++ b/lib/libsmb2.c
@@ -641,10 +641,6 @@ session_setup_cb(struct smb2_context *smb2, int status,
                 smb2->sign = 0;
         }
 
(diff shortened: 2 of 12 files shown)


lib/init.c
```diff
@@ -362,6 +362,13 @@ void smb2_destroy_context(struct smb2_context *smb2)
         free(discard_const(smb2->workstation));
         free(smb2->enc);
 
+        if ( NULL != smb2->connect_data )
```
sahlberg, line 365: Please match the style of the rest of the file.  The curly branch goes on the same line as the if-statement
, no spaces after (,  no space before ) and we don't compare != NULL. Instead this line should just be

if (smb2->connect_data) {

lib/libsmb2.c
```diff
@@ -526,6 +526,8 @@ free_c_data(struct smb2_context *smb2, struct connect_data *c_data)
         free(discard_const(c_data->share));
         free(discard_const(c_data->user));
         free(c_data);
+
+        if ( smb2->connect_data == c_data ) smb2->connect_data = NULL;  /* to prevent double-free in smb2_destroy_context */
```
sahlberg, line 530: We always use curly braces even if the body of the if-statement is just a single statement/expression, so this should be

```
if (smb2->connect_data == c_data) {
    smb2->connect_data = NULL;
}
```

sahlberg (commented):
You found a real memory leak. Just make these small style changes I suggested and it should be good.

lib/init.c
healytpk: done

lib/init.c
```diff
@@ -362,6 +362,13 @@ void smb2_destroy_context(struct smb2_context *smb2)
...
+        {
+            struct connect_data;                                                  /* defined in libsmb2.c            */
+            extern void free_c_data(struct smb2_context*, struct connect_data*);  /* defined in libsmb2.c            */
```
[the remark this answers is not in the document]
healytpk: done

sahlberg:
Since the first commit is not yet merged, lets combine these into a single commit instead.

In git you do that with "git rebase -i HEAD~2"
and then you change the second commit from "pick" to "f"  and save.
That will merge the two commits into a single commit instead.

The re-push to github    using "git push -f"   (-f since your local repo and githubs repo have diverged)

sahlberg:
I see you are struggling with updating the pull request. Don't worry about it. I can just cherry-pick the patch that you created tomorrow.

PR #277, opened 2023-11-13
Add Dreamcast (KallistiOS) build support

has207:
Tested with simple file opening a share and browsing files working on Dreamcast.

There's a small fixup for libsmb2 included to zero out pfd after it's initialized in sync.c, not sure why that works on other platforms but on dreamcast the contents of that memory are
always trashed in a way that makes consequent code think POLERR has been set.


sahlberg:
Merged. Thanks!
Maybe you can update the README with this support?

has207:
Will do!

sahlberg:
Why the sync.c code works on other platforms?  Nothing wrong with memset() here to clear it explicitely before the call to poll() but the only effect it has is that revents is cleared before the call to poll()

Maybe there is a bug in poll() on KallistiOS that does not clear the bits in pollfd.revents correctly?
Doesn;t matter,   using memset() to clear it is fine.

has207:
Yeah I suspect the other platforms poll() implementations may be more defensively programmed...I thought about adding the fix there but then figured it's good to have this structure be clean so added here in the end. Probably still worth updating KOS as well though

sahlberg:
I mentioned I had a prototype smb2 irx  but I never got it to be stable. It would work for a while and eventually hang and debugging iop code is hard.

If you want to test it out, here it is :  
https://github.com/sahlberg/ps2sdk/tree/smb2man

Wolf3s:
Nice pull request!!!

gyrovorbis:
Sorry to dig up an old PR, but GOOD WORK you guys, hell yeah!!! We are so happy to see this in the DC community!

Wolf3s:
> Sorry to dig up an old PR, but GOOD WORK you guys, hell yeah!!! We are so happy to see this in the DC community!

Thanks, i will see about add the project on kos-ports.

sahlberg:
> Sorry to dig up an old PR, but GOOD WORK you guys, hell yeah!!! We are so happy to see this in the DC community!

Never had a DC myself, me being a SONY guy. But I am willing to buy any console that can run my libsmb2 library.
As a newbie that knows nothing at all about DC, do you have pointers to where the scene is, what kind of HW I need to get etc etc  for me how to get started?  You can mail me directly at ronniesahlberg@gmail.com with suggestions.

has207:
This was pretty much my first DC contribution, but it was pretty painless. Nice community over at KallistiOS which provides a kind of posix programming layer for the DC and made this port really straightforward. There's a discord server linked off their main page. I've been looking more into low level programming on the DC since and trying to understand how the network game loading code works with the hopes of getting it to work over smb. (Spoiler: it seems like a real stretch if not totally impossible at this point, but I think some progress can be made so I'm still poking around). In terms of hardware it helps to have the broadband adapter, which is pretty pricey and hard to find, but it will let you load code over the network. Of course if you're trying to run any networking code on the DC it becomes a must have, but I believe a cheaper option for general coding is to use the serial port, which still requires additional purchasing on top of the base console. Other than that pretty much a controller with memory card and a cd burner is all I've needed so far.

Wolf3s:
@has207 and @gyrovorbis the port is been migrated to cmake.

Wolf3s:
So, now the building consists in use Makefile.platform instead of a bunch of makefiles on /lib directory

gyrovorbis:
Awesome, thank you! 

I'm going to see about getting this automatically built as a first-party library added to our `kos-ports` repo, so it's good to know that I should be using CMake there. Will report back if we run into any problems (or if we don't, and it all goes smoothly. lol)

Wolf3s:
Ok, here's the [link](https://github.com/Wolf3s/libsmb2/tree/cmake)   if you want to beta test.

PR #291, opened 2023-11-25
ESP_PLATFORM is sys/poll.h, not poll.h

nopnop2002:
ESP_PLATFORM has HAVE_POLL_H defined, but in sys/poll.h instead of poll.h.

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -45,7 +45,11 @@
 #endif
 
 #ifdef HAVE_POLL_H
+#ifdef ESP_PLATFORM
+#include <sys/poll.h>
+#else
 #include <poll.h>
+#endif
 #endif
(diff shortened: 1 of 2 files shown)


Wolf3s:
> HAVE_POLL_H



> ESP_PLATFORM has HAVE_POLL_H defined, but in sys/poll.h instead of poll.h.

Did you have esp sdk?

Wolf3s:
Hey i think you could comment `HAVE_POLL_H` in config.h finded in include\esp instead of doing:
`
#ifdef ESP_PLATFORM
#include <sys/poll.h>
#else
#include <poll.h>
#endif
`

Wolf3s:
Some platforms supports it

PR #295, opened 2023-12-01
Fixed "Implicit conversion loses integer precision" warning

amosavian:
This merge request changes varable types from `int` or `unsigned int` to platform dependent variables like `ssize_t` and `size_t`.

In few places where underlying SMB2 type is 32-bit length integer, an explicit type cast is added to silence the warning.

Also `smb2_timeval.tv_sec` became `time_t` in order to fix  2038 issue, and `smb2_timeval.tv_sec` to `suseconds_t` to make it generalized.


amosavian:
Please check this PR first. @sahlberg  @Wolf3s

Wolf3s:
Both compiled successfully, but this request i think it´s not good.

Wolf3s:
Fixing this warning will possibly lead to another problems.

amosavian:
My test cases in AMSMB2 run successfully. These changes should not lead to any issue as the the types are wider.
Also some variables that contain length are became unsigned. I think that would be logical error if they became negative and could lead to an exploit. @Wolf3s

Wolf3s:
> Few conversion warnings still remains in MSVC, I think we should fix them too @Wolf3s

Yes it should, mainly on the x64 part.

Wolf3s:
> My test cases in AMSMB2 run successfully. These changes should not lead to any issue as the the types are wider. Also some variables that contain length are became unsigned. I think that would be logical error if they became negative and could lead to an exploit. @Wolf3s

So there is no exploit or logical error in this case.

Wolf3s:
Plans to continue Or you will leave it?

PR #300, opened 2023-12-18
Add Xbox/Xbox 360 Support.

Wolf3s:
Note:
The 360 port was made by @brentdc-nz based on: https://github.com/brentdc-nz/XBMC-360/tree/master/xbmc360/libraries/libsmb2
and for the original xbox it was me of course.


include/smb2/libsmb2.h
```diff
@@ -76,13 +76,22 @@ struct smb2dirent {
...
+#if defined(_WINDOWS)
 #include <winsock2.h>
+#elif defined(_XBOX)
+#include <xtl.h> 
+#include <winsockx.h>
+#endif
+
+#if defined(_WINDOWS) || defined(_XBOX)
+typedef SOCKET t_socket;
+#elif defined(_XBOX)
```
sahlberg, line 88: This logic seems odd.  
Should line 86 be "#if defined(_WINDOWS) && !defined(_XBOX)"
or something?

lib/aes.c
```diff
@@ -459,6 +459,9 @@ void AES128_ECB_decrypt(uint8_t* input, const uint8_t* key, uint8_t *output)
 {
   // The array that stores the round keys.
   uint8_t roundKey[176];
+	
+  // Copy input to output, and work in-memory on output
+  BlockCopy(output, input);
```
sahlberg, line 464: We already do this on line 467

lib/init.c
```diff
@@ -279,9 +279,11 @@ struct smb2_context *smb2_init_context(void)
         char buf[1024] _U_;
         int i, ret;
         static int ctr;
-
+#ifdef _XBOX
+        srandom(time(NULL) ^ 1 ^ ctr++);
```
sahlberg, line 283: Can probably get rid of ^1 as it is a constant change and does not improve randomness.
I.e.
srandom(time(NULL) ^ ctr++)

lib/socket.c
```diff
@@ -883,13 +895,15 @@ connect_async_ai(struct smb2_context *smb2, const struct addrinfo *ai, int *fd_o
                 ((struct sockaddr_in *)&ss)->sin_len = socksize;
 #endif
                 break;
+#ifdef ENABLE_IPV6
```
sahlberg, line 898: AF_INET6 should be available by default on most platforms, so instead of opting-in to have ipv6
can we change this to opt-out of ipv6 and only use it on those platforms that lack ipv6?

I.e. a #ifdef DISABLE_IPV6 instead

lib/socket.c
```diff
@@ -1050,12 +1078,17 @@ smb2_connect_async(struct smb2_context *smb2, const char *server,
...
-#ifdef _WINDOWS
+#if defined(_WINDOWS) || defined(_XBOX)
                 if (err == WSANOTINITIALISED)
                 {
-                        smb2_set_error(smb2, "Winsock was not initialized. "
+#ifdef _XBOX
+  					smb2_set_error(smb2, "Winsockx was not initialized. "
+                                "Please call WSAStartup().");                      
+#else
+					smb2_set_error(smb2, "Winsock was not initialized. "
```
sahlberg, line 1088: Lets not have a big block of code and ifdefs just to change an error string that the user will likely not see anyway.
So get rid of this change completely.

We should have as little #if defined(platform)   as possible in the general *.c files as possible

lib/xbox/emu_socket.cpp
```diff
@@ -0,0 +1,623 @@
+/*
+ *      Copyright (C) 2005-2013 Team XBMC
+ *      http://xbmc.org
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
```
sahlberg, line 8: This is a showstopper. This file is GPLv2 while libsmb2 is LGPLv2.1.
This change would change the licence of libsmb2 and is thus a showstopper.

We would need to ask the kodi folks to re-licence this file as LGPL before we can use it.
Or we would have to re-write it.

There are I think two additional files that are GPLv2 too that needs to be re-licenced or replaced.

lib/errors.c
```diff
@@ -1107,32 +1107,38 @@ int nterror_to_errno(uint32_t status) {
...
         case SMB2_STATUS_INVALID_DEVICE_REQUEST:
+
         case SMB2_STATUS_ILLEGAL_FUNCTION:
-        case SMB2_STATUS_INVALID_PARAMETER:
+
+		case SMB2_STATUS_INVALID_PARAMETER:
         case SMB2_STATUS_NOT_SUPPORTED:
         case SMB2_STATUS_NOT_A_REPARSE_POINT:
                 return EINVAL;
+#ifndef XBOX_PLATFORM
```
sahlberg, line 1125: Instead of adding #ifdefs here in this file,
can we just define SMB2_STATUS_STOPPED_ON_SYMLINK to be some random value in the xbox compatibility header files?

That is usually how I deal with "define is missing on platform foo", I add a dummy #define foo whatever to the compatibility files.
This is all to avoid having platform specific #ifdefs int he main code.

sahlberg (changes requested):
Most are nitpicks  except for the license incompatibility.
We can not put GPLv2 inside the main library.   GPLv2 is fine in the test/exaple programs that link with libsmb2 but not in libsmb2 itself.

include/smb2/libsmb2.h
Wolf3s: Yep there's some mistakes

lib/socket.c
Wolf3s: On amiga os port looks more effective than Xbox 360 port.

lib/xbox/emu_socket.cpp
Wolf3s: Well like i said above i have to merge these guys but using LGPLv2.1 code

Wolf3s: Seeing here some code are not GPL :+1:  At least but i will probally Join these non GPL code into one and use mvcs gethost instead of xbmc ones.

Wolf3s:
All of these request i will try tommorow and remove.

lib/aes.c
Wolf3s: Solved.

include/smb2/libsmb2.h
Wolf3s: Solved.

lib/init.c
Wolf3s: Solved using the original code

Wolf3s:
https://github.com/sahlberg/libsmb2/pull/300/commits/41443f10134647bd49ce9f1ea4b82d516e5eea21

lib/errors.c
Wolf3s: Solved.

lib/xbox/emu_socket.cpp
Wolf3s: Removed the XBMC socket emulation and using wide code instead.

Wolf3s:
After the removal of gpled code.

lib/init.c
```diff
@@ -287,9 +289,11 @@ struct smb2_context *smb2_init_context(void)
         char buf[1024] _U_;
         int i, ret;
         static int ctr;
-
+#ifdef _XBOX
+        srandom(time(NULL) ^ ctr++);
+#else
         srandom(time(NULL) ^ getpid() ^ ctr++);
-
```
sahlberg, line 292: There are already too many #ifdef _platform_ in the lib/*.c part of the library
and on later thought I think maybe better would have been to add a :
int getpid() { return 0}
to the xbox part of the compat.c/h

But It is fine, we can do those cleanups later.  I have myself also added these kinds of ifdefs which I now consider a mistake.

Wolf3s: I think this one is not necessary to add getpid() but it's a good idea tough i was trying this but using `#define getpid()`

Wolf3s: Soon i will add patch(more portable way) of getpid.

sahlberg: Are you letting the xbox folks know that official support in upstream is now available and they won't need to carry an out-of-tree fork any more?

Wolf3s: > Are you letting the xbox folks know that official support in upstream is now available and they won't need to carry an out-of-tree fork any more?

Yes, i contacted in forums about xbox support and they stayed well about the news. 
I will send link of emuextras(which is the official place of the xbox community);
https://www.emuxtras.net/forum/viewtopic.php?f=186&t=7743
https://www.emuxtras.net/forum/viewtopic.php?f=229&t=7744

Wolf3s: https://github.com/sahlberg/libsmb2/pull/305

Wolf3s: Getpid fixes

PR #280, opened 2023-11-14
Support Clang/Xcode and support Kerberos on macOS

amosavian:
This pull request adds module.modulemap to allow clang to treat this library as
a framework, allowing use of "`import`" rather than include and provides better
support for Swift binding.

Also krb5wrapper has been modified to support macOS.
I've disabled a section of code as `gss_set_neg_mechs()` is not defined on macOS.

Please note I've changed two cases of angled-bracket include to double-quoted one
as it did not compile on my computer. **I hope it won't break anything on other systems.**

PR #301, opened 2023-12-22
Add Support for Clang/Xcode and Kerberos on macOS 

Wolf3s:
Fixed branch from #280


include/smb2/smb2.h
```diff
@@ -19,7 +19,11 @@
 #ifndef _SMB2_H_
 #define _SMB2_H_
 
+#ifdef __APPLE__
+#include <smb2-errors.h>
+#else
 #include <smb2/smb2-errors.h>
+#endif
```
sahlberg, line 26: This seems like something wrong with the setup for the apple compiler. It should use both -I include and -I include/smb2
so that smb2-errors.h should always be found regardless of if <smb2-errors.h> or <smb2/smb2-errors.h> are used.

Wolf3s: Apple xcode normally would have some stuff like including in one dir.

PR #326, opened 2024-03-28
PS3: fix uninitialize addrinfo struct causing crash

cy33hc:
Fix crash caused by uninitialized addrinfo struct especially in for loop in socket.c that references ai->ai_next. This has been broken since commit 3cbcc5a32af2183fb28864be1bd7907325c0b81d


sahlberg:
Thanks for this fix!

PR #327, opened 2024-03-28
Cmake and windows socket check fixes.

memecode:
These fixes are related to fixing the cmake target includes for the smb2 target. And the socket validity checks on windows, where the SOCKET (t_socket) type is unsigned.


sahlberg:
Merged but the windows build is broken.
I fixed the non-windows builds by moving the defines so they would apply also for non xbox/windows/mingw builds
so all other platforms should work now. But windows still fails.

Can you have a look?

memecode:
@sahlberg I've committed a windows build fix on the source branch: [00fd58666f7005d40cafc80f55e1b660069bca21](https://github.com/sahlberg/libsmb2/compare/master...memecode:libsmb2:all-fixes)

sahlberg:
Awesome. Thanks!
Send a pull request and I will merge it.

PR #328, opened 2024-03-29
Windows build fixes.

memecode:
Moved the winsock header includes above the place that tries to use the SOCKET symbol.


sahlberg:
Merged. Thanks!

PR #334, opened 2024-04-30
Fix build for xbox

Wolf3s:
xtl header is equivalent of windows.h
More xbox warning fixes.

### DIFF
--- a/lib/smb2-cmd-query-directory.c
+++ b/lib/smb2-cmd-query-directory.c
@@ -39,6 +39,14 @@
 #include <stddef.h>
 #endif
 
+#ifdef HAVE_TIME_H
+#include <time.h>
+#endif
+
+#ifdef HAVE_SYS_TIME_H
+#include <sys/time.h>
+#endif
+
 #include <errno.h>
 
 #include "compat.h"
--- a/lib/smb2-cmd-query-directory.c
+++ b/lib/smb2-cmd-query-directory.c
@@ -79,16 +79,16 @@ smb2_decode_fileidfulldirectoryinformation(
         fs->name = utf16_to_utf8((uint16_t *)&vec->buf[80], name_len / 2);
 
         smb2_get_uint64(vec, 8, &t);
-        win_to_timeval(t, &fs->creation_time);
+        win_to_timeval((time_t)t, &fs->creation_time);
 
         smb2_get_uint64(vec, 16, &t);
-        win_to_timeval(t, &fs->last_access_time);
+        win_to_timeval((time_t)t, &fs->last_access_time);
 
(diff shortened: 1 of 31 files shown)


lib/smb2-cmd-query-directory.c
```diff
@@ -54,7 +54,7 @@ smb2_decode_fileidfulldirectoryinformation(
     struct smb2_iovec *vec)
 {
         uint32_t name_len;
-        uint64_t t;
+        time_t t;
```
sahlberg, line 57: We can't do this. time_t is not guaranteed to be 64 bit on all platforms and would cause an out of bounds write later in the call to smb2_get_uint64().

Revert this change.

lib/smb2-data-file-info.c
```diff
@@ -51,18 +51,18 @@ smb2_decode_file_basic_info(struct smb2_context *smb2,
                             struct smb2_file_basic_info *fs,
                             struct smb2_iovec *vec)
 {
-        uint64_t t;
+        time_t t;
```
sahlberg, line 54: same as above

lib/smb2-data-filesystem-info.c
```diff
@@ -51,10 +51,10 @@ smb2_decode_file_fs_volume_info(struct smb2_context *smb2,
                                 struct smb2_file_fs_volume_info *fs,
                                 struct smb2_iovec *vec)
 {
-        uint64_t t;
+        time_t t;
```
sahlberg, line 54: same as above

sahlberg (changes requested):

lib/smb2-cmd-query-directory.c
Wolf3s: 👍

lib/socket.c
```diff
@@ -303,7 +303,7 @@ smb2_write_to_socket(struct smb2_context *smb2)
         return 0;
 }
 
-typedef ssize_t (*read_func)(struct smb2_context *smb2,
+typedef size_t (*read_func)(struct smb2_context *smb2,
```
sahlberg, line 306: I don't think we can do this change from ssize_t to size_t.
The helper functions are based on how functions like readv() returns error (<0)
and smb2_read_data() depends on this.
Line 351 in socket.c , after we hare read the data via these functions, smb2_read_data() depends on them returning <0.

Leave these function returning ssize_t for now, that way the mimic posix functiions for nonblocking i/o. return >0 for number of bytes read, <0 for check errno, buffer empty and try again later   or ==0 for end of file

Wolf3s: Ok, i will leave it, btw i changed also some variables that takes size_t instead of ssize_t.

PR #329, opened 2024-03-30
Add smb2_ in some private smb2 functions and _SOCKET for portability

Wolf3s:
This pr is based on @cy33hc nintendo´s switch branch, since devkitpro and possibly other´s sdk´s has similar functions.

PR #337, opened 2024-05-12
Add support for Nintendo Switch And Nintendo 3DS

Wolf3s:
It depends: #334 and #329


Wolf3s:
Port by @cy33hc for Nintendo Switch and myself with Nintendo 3DS.

PR #341, opened 2024-05-19
Update to latest PS2SDK.

Wolf3s:


sahlberg:
Very nice!  Getting rid of a whole bunch of platform specific ifdefs.

PR #365, opened 2024-09-12
implement server side, netprot works basically

bdodge:
This is a prototype implementation of serving SMB


sahlberg:
Nice.
I activated a build for this but some targets fail.
You can ignore the consoles that fail, I will fix that up, but can you take a look at the failed windows build and once that works I will merge this.

sahlberg:
Ah, it is this structure.  struct  validate_neg_info {
Not sure if all platforms support the packed attribute.

You shouldn't assign straight into this structure since that breaks big-endian systems.
Assing using the htole*() calls instead.

sahlberg:
Merged, thanks. I will sort out the windows/cmake issues and later will fix the game console builds.

This is great.  I will find time and build it as a smb2 server for ps2.

bdodge:
cool.

I made a lot more progress the last few days too.  I can serve Linux
(nautilus), Linux (samba) and Mac OSX (netbsd, probably samba) on SMB 2.1.
There were a bunch of issues in my asn1 parsing / generation I fixed and
now I detect if spnego wrapping is needed or not and auto-add if needed.

Started to enable SMB311 now to try and get my windows 11-pro machine to
connect.   It doesn't work with samba smbclient yet,,,  I get a signature
mismatch.  It seems it wants a signature of the session-response pdu? If I
hack smb-signing to allow pdu signatures for session commands if the
server-redir flag is set in the header I do get a signature in the header
and the signed flag set, but smbclient complains the signature doesn't
match its calculation.  I suppose the signing key isn't being made
properly?  I also havent tried to work out pre-auth stuff yet since that
hurts my head..

Also, a real bonafide bug in smb2_encode_preauth_context(struct
smb2_context *smb2, struct smb2_pdu *pdu).  You are padding data_len to 64
bits but samba insists that the data_len is equal to the salt length (32) +
6.  Its a silly restriction in samba but I made data_len 38 and padded the
alloc len up and samba is happy now.

Any hints on getting smb311 working would be great.  i am not clear at all
if the signing key is being  created properly (I just used the same code
you do for the client side) which works for 2.1 at least.

finally, the packed struct hack was a short-cut.  I was just in a hurry.
the whole ioctl stuff needs to be passed to the server-handler function
anyway, and the structs properly packed and unpacket in the cmd handler..
just got tedious and I wanted to test.  But this brings up a big can of
worms:  Try compiling for Arm based macs and you get a million alignment
warnings, starting in alloc.c with "container_of" type stuff.  Not sure it
is fixable.  I did get it all working on my x86 mac though suppressing
those.

Cheers, Brian

sahlberg:
SMB3.1.1 is a little bit different when generating the keys. It adds
protection also of the initial NegotiateProtocol packets before
authentication has completed and feeds a preauthentication hash into the
key generator that covers also the initial negotiate protocol packets.

The code in libsmb2 SHOULD generate this correctly but I don't remember if
I actually tested it.
You need to negotiate a SMB2_PREAUTH_INTEGRITY_CAPABILITIES (sha512) asa
negotiate protocol context.
Then update and compute a rolling hash of all the packets until the key is
generated.
See smb3_update_preauth_hash in libsmb2.c how this is done.

You can also print the hashes as you compute them for the received packet
and then for the reply you send
and capture this with wireshark.  Wireshark will show what the hash should
be at each step.
We added this to make debugging easier when we added 3.1.1 to the linux
client.

In wireshark sources you can see how wireshark does it in
epan/dissectors/packet-smb2.c

sahlberg:
"Any hints on getting smb311 working would be great.  i am not clear at all
if the signing key is being  created properly (I just used the same code
you do for the client side) which works for 2.1 at least."

In addition to using wireshark to make sure you compute the preauth hash correctly for each packet
there is also ways to verify you generate the session key correctly.

The linux kernel cifs client can dump the session keys (so that you can paste them into wireshark to decrypt the stream).
You need to build the kernel module   CONFIG_CIFS_DEBUG_DUMP_KEYS=y
and then once a client connect it will write the keys for that connection into dmesg.

That way you can compare the key you generate server-side with the corresponding key that linux cifs client generates clientside.

PR #369, opened 2024-09-27
Xbox build fixes with C89 compat code.

Wolf3s:


Wolf3s:
@sahlberg i fixed the amiga os build.

PR #370, opened 2024-09-29
Bdd/pdu passthrough

bdodge:
adds the ability to pass blobs along as blobs (not packing/unpacking) to be able to proxy complex pdus like query-info, etc.
adds the ability to get/set tree-id for pdus by application
uses waitqueue for requests to allow setting reply message-id to match the request for server
remove waitqueue use for replys as they are never removed
adds handing of "set-notify" command/reply
fixes issues with credits as a server (kind of maybe)


sahlberg:
Merged, thanks!

PR #371, opened 2024-10-01
Move modulemap and remove Apple special imports

amosavian:
This PR removes hacky `__APPLE__` directive imports as it not necessary anymore by moving `module.modulemap` file to `include` directory.

Also it fixes a header and implementation definition of `smb2_cmd_set_info_reply_async` method discrepancy which caused a build error.


sahlberg:
Merged, thanks

PR #373, opened 2024-10-03
oplock and lease break pdu handling

bdodge:
Adds basic oplock-break and lease-break handling.  They are a bit special in that they use the same command code but have different stuct sizes, so the "fixed" part of the pdu is only the 2 byte struct size and the "variable" part is used to read in the rest of the pdu depending on sub-type.

Tested and working


sahlberg:
Merged, thanks!  Sorry for the delay to review and merge.

PR #374, opened 2024-10-06
Fix pwrite offset encoding

amosavian:
This PR fixes an issue in write and pwrite functions, due to missing offset.

This is a critical issue in ftruncate, which I could not find the isssue yet.


sahlberg:
merged, thanks!

Issue #377, opened 2024-10-12
Memory deallocation question


erikbernsen:
I'm running into a memory issue in AMSMB2 library that is a kind of wrapper around libsmb2. See https://github.com/amosavian/AMSMB2/issues/116 

The issue is that in the following location libsmb2 allocates memory, but it never gets deallocated.
<img width="1042" alt="leaks" src="https://github.com/user-attachments/assets/f9d1a3f3-2b7f-4e4c-a654-70371fa0b543">
Initially I thought this should get deallocated in smb2_free_pdu, but then realised since payload is just a void pointer in smb2_pdu it cannot know how to deallocate.

My question is this: who is responsible for freeing the memory allocated in smb2_process_read_variable? It is the libsmb2 library, or is it the user of the library?

erikbernsen:
I observed that the same memory leak issue happens in the example smb2-cat-async. And I figured this is a memory allocation issue with libsmb2, not a mistake in AMSMB2 wrapper. I created a PR that should fix this issue.

PR #378, opened 2024-10-13
Fix a memory leak.

erikbernsen:
Memory allocated in file smb2-cmd-read.c and function smb2_process_read_variable is never deallocated. This PR fixes that issue. Since pdu stores a void* pointer to the payload it doesn't know how to deallocate it. I added a callback to struct smb2_pdu that frees the payload and is set in the smb2_process_read_variable function

### DIFF
--- a/include/libsmb2-private.h
+++ b/include/libsmb2-private.h
@@ -258,6 +258,12 @@ struct smb2_context {
         struct smb2_context *next;
 };
 
+/*
+ * Callback for freeing a payload.
+ */
+typedef void (*smb2_free_payload)(struct smb2_context *smb2, void *payload);
+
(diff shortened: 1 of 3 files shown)


erikbernsen:
It's a fix to this issue: https://github.com/sahlberg/libsmb2/issues/377

erikbernsen:
Of course, libsmb2 is a completely new codebase for me. It's a sample way of solving the memory leak, but maybe someone with more in depth knowledge of the code base has a better solution.

sahlberg:
Thanks. Merged.

I have started adding tests and I added a valgrind test for this as test_0301_...

Thanks again!

PR #380, opened 2024-11-28
Implementation of Change Notify Request client side

tryggvil:
This implements both sync and asynchronous library calls for Change Notify.

The sync version smb2_file_notify_change_information does a one-off request to a notify change
The async version smb2_notify_change_async has the option of doing a new request on completion of the previous notify change call, thus listening for incoming changes.

There is one change to smb2_open to allow it to call it with O_DIRECTORY to be able to get the fileid for a directory.


examples/smb2-notify.c
```diff
@@ -0,0 +1,183 @@
+/* -*-  mode:c; tab-width:8; c-basic-offset:8; indent-tabs-mode:nil;  -*- */
+/*
+   Copyright (C) 2017 by Ronnie Sahlberg <ronniesahlberg@gmail.com>
```
sahlberg, line 3: Change this

include/smb2/smb2.h
```diff
@@ -1014,18 +1014,18 @@ struct  smb2_ioctl_validate_negotiate_info {
         uint16_t dialect;
 };
 
-#define SMB2_CHANGE_NOTIIFY_FILE_NOTIFY_CHANGE_FILE_NAME    0x00000001
```
sahlberg, line 1017: This whole change should be folded into the previous commit.

sahlberg (changes requested):

lib/libsmb2.c
```diff
@@ -2791,6 +2796,145 @@ smb2_oplock_break_notify(struct smb2_context *smb2, int status, void *command_da
...
+                        -nterror_to_errno(notify_change_data->status),
+                        fnc,
+                        notify_change_data->cb_data
+                );
+        }
+        if (notify_change_data->loop) {
+                smb2_notify_change_filehandle_async(smb2, notify_change_data->fh, notify_change_data->flags, notify_change_data->filter, 
+                        notify_change_data->loop, notify_change_data->cb, notify_change_data->cb_data);
+        } else {
+                smb2_close(smb2, notify_change_data->fh);
```
amosavian, line 2879: I believe calling `smb2_notify_change_filehandle_async` should not result in closed file handle, whule it is reasonable to be closed if `smb2_notify_change_async` is called.
Can we put a flag in command data for this case?

examples/smb2-notify.c
tryggvil: I have updated this

lib/libsmb2.c
tryggvil: smb2_close is only called if notify_change_data->loop is set to false, which means it is supposed to be a one-off operation, and in that case it should be normal to close the file handle, as it is opened by smb2_notify_change_async

amosavian: I can call `smb2_notify_change_filehandle_async` with loop = false with a filehandle created by myself and this method should not close the handle in this case IMO.

sahlberg: I agree with amosavian here. The function should not close the handle but I understand why it does.
The smb2_notify_change_async() function is just a small wrapper that takes a path and converts it to a filehandle and then calls the actual notify function, which takes a file handle.

Probably get rid of smb2_notify_change_async() and thus also the reason for the loop variable or closing the handle in the callback.
And then rename smb2_notify_change_filehandle_async() to smb2_notify_change_async().

amosavian: Another point I think that should be addressed is to check if the filehandle is still valid and open before calling `smb2_notify_change_filehandle_async` again.
@tryggvil @sahlberg

lib/libsmb2.c
```diff
@@ -1213,8 +1213,13 @@ smb2_open_async_with_oplock_or_lease(struct smb2_context *smb2, const char *path
                         break;
         }
 
-        /* create options */
-        create_options |= SMB2_FILE_NON_DIRECTORY_FILE;
+        if (flags & O_DIRECTORY) {
```
sahlberg, line 1216: O_DIRECTORY is a linux only thing.
That is not a super big deal but we will have to add some

#ifndef O_DIRECTORY
#define O_DIRECTORY <some value that does not clash with any of the other flags on linux/osx/windows>
#endif 

to this file

sahlberg (commented):
Please fix the O_DIRECTORY issue, perhaps like how I suggested.

stefbon:
Hi,

you write:

"Another point I think that should be addressed is to check if the
filehandle is still valid and open before calling
smb2_notify_change_filehandle_async again."

Isn't it so that if it is not valid anymore you would receive an error
from the system? So you do not have to check it's valid.

S. Bon

sahlberg:
Maybe don't need to check if it is valid. 
That should be the callers responsibility. Don't call this function if the handle is not valid.  
Just like the caller is responsible to not call smb2_read and other functions with an invalid handle.

And if it is invalid, the server will return invalid-parameter so it should be fine.

sahlberg:
Thanks for the contribution. I folded some commits together and have merged them to my master branch.

I fixed the O_DIRECTORY issue and also removed some calls to [f]printf since we can not assume these can be used.

sahlberg:
Manually merged. Thanks!

PR #381, opened 2024-12-10
implement kerberos auth for server (and proxy) including constrained delegation

bdodge:
This provides for krb5 authentication to libsmb2 server and can co-exist with local ntlmssp auth in the case krb5 ntlmssp isnt installed. TODO - detect if krb5 can handle ntlmssp and offload to krb5 in that case.  Need to figure a way to ask krb5 lib if it can do that.

Also provides for constrained delegation in proxy use-case where the original client credentials can be used to get a proxy credential to use for proxy client to actual server.

Adds a "suppress_errors" flag to ntlmssp message type sniffing to allow for auto-detect of auth method during negotiation by using the get-message-type function to determine if a valid ntmssp message is in a blob


sahlberg:
Thanks.
I have merged these three commits into a single one and added to master : 

fb71126ea41b5c8886c74638c4925957413c5837 proper type for cred handle
2e416aa1f491ff84f63814dc773f3b296cf0e582 add krb5 cred handle passing api
c0e8924588ed1cc8724290dd3e4a707e320b7aaa add krb5 cred handle passing and release mechanism for proxying

since they belong together. Please rebase ontop of master and remove these three commits from your push request.

bdodge:
It took me a while to merge/sync.  I had t make a few merge fix commits

bdodge:
Also, there are a few more fixes added.  I now properly pass-through
STATUS_PENDING messages, and fixed a compile error in an example

lib/smb2-cmd-write.c
```diff
@@ -262,10 +262,11 @@ smb2_process_write_request_fixed(struct smb2_context *smb2,
         smb2_get_uint32(iov, 4, &req->length);
         smb2_get_uint64(iov, 8, &req->offset);
         memcpy(req->file_id, iov->buf + 16, SMB2_FD_SIZE);
-        smb2_get_uint32(iov, 24, &req->channel);
-        smb2_get_uint32(iov, 28, &req->remaining_bytes);
-        smb2_get_uint16(iov, 32, &req->write_channel_info_offset);
-        smb2_get_uint16(iov, 34, &req->write_channel_info_length);
+        smb2_get_uint32(iov, 32, &req->channel);
```
sahlberg, line 265: The offsets for write_channel_info_offset, info_length and flags seems wrong. The look to be 2 too high. Same above.

Can you check this?

sahlberg:
I manually merged virtually all of this trying to make the git history look nice. 
But then it started diverging so I slapped in most of the krb5 and ntlm stuff as a single big patch.

You can check the delta between your branch and master by doing a git diff sahlberg/libsmb2.
It should pretty much all be just whitespace changes that I skipped  except for the changes to
the write command. See the commend above that the offsets seems to be wrong.

lib/smb2-cmd-write.c
brian-level: Great catch.. they were even more wrong before this change which kind of worked, but now they should be correct

lib/krb5-wrapper.c
```diff
@@ -601,11 +601,11 @@ krb5_init_server_cred(struct smb2_server *server, struct smb2_context *smb2, con
...
 #ifndef __APPLE__
-                /* get a proxy ticket when user is known in session reply.  note that this
-                 * isnt absolutely required if the accept returns a delegatable cred but it
-                 * does help find issues
+                /* set this non-0 to also get proxy ticket when user is known in
+                 * session reply.  note that this isnt required if the accept returns
+                 * a delegatable cred but it might help find issues (or cause some)
                  */
-                auth_data->s4u2self = 1;
+                auth_data->s4u2self = 0;
```
brian-level, line 608: Note - if a delegatable cred is gotten there is no need to also do an s4u2self request.  In fact, the code I made to do that was a bit wrong, and ended up doing a service -> remote-service check which worked by mistake on my AD DC because of extra delegation I had setup.  It wants to a user->remote-service check instead which I will work on at some point so plz leave the code in here?

lib/libsmb2.c
```diff
@@ -3206,12 +3206,10 @@ smb2_query_directory_request_cb(struct smb2_server *server, struct smb2_context
                         pdu = smb2_cmd_query_directory_reply_async(smb2, req, &rep, NULL, cb_data);
                 }
         }
+        if (req->name) {
+                smb2_free_data(smb2, discard_const(req->name));
+        }
         if (pdu != NULL) {
```
brian-level, line 3212: CHANGE Note: the comment was wrong.. this needs freeing always.  I had patched this to work around a use-after-free issue elsewhere, but that is the fault of elsewhere, not this lib.

lib/libsmb2.c
```diff
@@ -3319,7 +3317,7 @@ smb2_general_client_request_cb(struct smb2_context *smb2, int status, void *comm
                 smb2_close_context(smb2);
                 return;
         }
-        if (status == SMB2_STATUS_CANCELLED) {
+        if (status == SMB2_STATUS_CANCELLED || status == SMB2_STATUS_SHUTDOWN) {
```
brian-level, line 3320: CHANGE Note: Important one, the server side will now crash on smb2 context destroy if this change isn't included

Issue #384, opened 2024-12-16
[cmake] cannot disable krb5 or GSSAPI


michael-doubez:
I need to compile without krb5 or gssapi.  I tried using `-DCMAKE_DISABLE_FIND_PACKAGE_GSSAPI=ON` and `-DCMAKE_DISABLE_FIND_PACKAGE_LibKrb5=ON`. It is working but the source code is still expecting the libraries.

The reason seems to be that `cmake/ConfigureChecks.cmake` detects the `krb5.h`header and defines `HAVE_LIBKRB5`.

Could you please add `SMB2_KRB5_ENABLE` and `SMB2_GSSAPI_ENABLE` options ?
And completely disable related `find()` and `HAVE_`.

I am also in the process of updating the vcpk recipe.
Could you please tag it once it is done ?

If you prefer, I can contribute a change.

sahlberg:
Please if you could send a pull request.
I am not very familiar with CMake (I am an autotools man)

michael-doubez:
Thank you
https://github.com/microsoft/vcpkg/pull/42779

PR #387, opened 2024-12-17
cmake: add ENABLE_LIBKRB5 option

tguillem:
On by default, but the user can now disable it (like configure.ac).


tguillem:
Wait before merge, will do the same for GSSAPI, cf. https://github.com/sahlberg/libsmb2/issues/384

sahlberg:
Merged. Thanks!

PR #390, opened 2025-01-03
Allow use of non-default keytab for server service authentication

bdodge:
by specifying a key table file path, the kerberos credentials for the server service can be acquired without the need for a password or doing"kinit" in the shell and running as root to allow access to the krb5 default keytab

The programmatically does the equivalent of 

kinit -kt <keytab path> -c <memory-cache> cifs/<hostname>

getting initial credentials from the KDC using the key table

Then to acquire cred to accept an client connection, gss_acquire_cred_from is used with the keytab and cache path instead of the defaults.  Works great.

Also adds a daily renew of credentials to the server can run continuously in kerberos world

former behavior happens if no keytab path set in server context

Benefits
- no password needed to pass to or store in server
- no use of the default cache, so clients and other programs can kinit on the same machine without affecting the server
- no need to run as root user (to get access to krb5.keytab for initiator creds)
- no need to merge the key for the service in to the default keytab, can just use standalone

### DIFF
--- a/include/smb2/libsmb2.h
+++ b/include/smb2/libsmb2.h
@@ -24,7 +24,7 @@ extern "C" {
 #endif
 
 #define LIBSMB2_SHARE_ENUM_V2 1
-        
+
 struct smb2_iovec {
         uint8_t *buf;
         size_t len;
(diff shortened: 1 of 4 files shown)


sahlberg:
Thanks!
Manually folded into a single commit and merged.

When you have a commit followed by a small fix to the previous commit
you should "git rebase -i HEAD~3" and use "fold" ro merge them into a single commit followed by a forced push to update the pull request.
That makes the git history look nicer.

(I also changed #if HAVE_LIBKRB5 to #ifdef HAVE_LIBKRB5 as the former does not work for CMake.)

bdodge:
Your git-foo is much stronger than mine!

I think I will leave the client/krb5 path alone nnow.  Any direct client
use of a keytab is probably a security issue and shouldn't be encouraged
when they can just kinit / use-password

Cheers,  Brian

PR #407, opened 2025-03-19
Fix building on OpenBSD

brad0:
OpenBSD does not have the ENOLINK errno.


brad0:
I am not sure what errno to use. I made a patch and have used ENOATTR for now and it builds.

sahlberg:
ENOATTR is as good as any other.

Merged, thanks!

PR #411, opened 2025-04-19
Fix smb2_echo failing even if socket fd is valid

cy33hc:
As in title, the smb2_echo fails if socket fd is valid. This has being broken since PR #327


sahlberg:
Merged. Thanks!

PR #415, opened 2025-06-03
Adapt portable-endian.h for GNU/Hurd

pinotree:
Use `<endian.h>` & `<byteswap.h>`, as available from GNU libc.


sahlberg:
Merged, thanks!

PR #418, opened 2025-06-14
fix target info field offset for non-litle endian systems

piru:
Corrected the wrong size byteswap function being used when constructing the target info field offset - it only worked on little endian systems that have no-op "host to le" functions.


sahlberg:
Good catch. thanks!

PR #422, opened 2025-07-29
fix installation readme nintendo switch, nintendo ds, nintendo wii-u …

Wolf3s:
…and Dreamcast.


Wolf3s:
> Merged, thanks!
No problem.

PR #453, opened 2026-03-23
Update Dreamcast port

pcercuei:
- The CMAKE_SYSTEM_NAME is defined as lowercase 'dreamcast' by KallistiOS' CMake scripts.
- Added some glue code to allow mapping a SMB server into KallistiOS' virtual filesystem. Users now just need to call kos_smb_init() with a server URL as the argument, to have it mounted as /smb in the VFS.


sahlberg:
> > Merged. I checked it in but wonder if vfs.c/h shouldn't rather be in the sdk for kallistoos instead of libsmb2.
> 
> > > Merged. I checked it in but wonder if vfs.c/h shouldn't rather be in the sdk for kallistoos instead of libsmb2.
> > 
> > 
> > Up to you, if you prefer we can have it in KallistiOS directly.
> 
> We have vfs aswell for ps2, but it's wLaunchelf. i'm implementing for a fork of [OPL](https://github.com/NathanNeurotic/wOPL/tree/smb2/3/src) It should be awesome port this guy to other platforms.

I have no objections.
If someone wants to make this into a general wrapper for embedded systems, I will not object to a PR.

Wolf3s:
> > > Merged. I checked it in but wonder if vfs.c/h shouldn't rather be in the sdk for kallistoos instead of libsmb2.
> > 
> > 
> > > > Merged. I checked it in but wonder if vfs.c/h shouldn't rather be in the sdk for kallistoos instead of libsmb2.
> > > 
> > > 
> > > Up to you, if you prefer we can have it in KallistiOS directly.
> > 
> > 
> > We have vfs aswell for ps2, but it's wLaunchelf. i'm implementing for a fork of [OPL](https://github.com/NathanNeurotic/wOPL/tree/smb2/3/src) It should be awesome port this guy to other platforms.
> 
> I have no objections. If someone wants to make this into a general wrapper for embedded systems, I will not object to a PR.

It should be more easier than implementing usmb2 to opl for streaming games, i will take a look at it, when i have some time.

PR #460, opened 2026-07-07
Add SET_INFO/SMB2_0_INFO_SECURITY encoding support and a set-security-info example

RidaShamasneh:
smb2_encode_set_info_request() (lib/smb2-cmd-set-info.c) previously only handled SMB2_0_INFO_FILE, so any attempt to set a file's security descriptor (DACL) via SET_INFO failed client-side before ever reaching the wire. Add smb2_encode_security_descriptor()/ smb2_security_descriptor_size() (lib/smb2-data-security-descriptor.c), mirroring the existing smb2_decode_security_descriptor() (same SID/ACE/ ACL coverage), and wire it into the SMB2_0_INFO_SECURITY case of smb2_encode_set_info_request().

Flesh out examples/smb2-set-security-info.c (previously an unused stub) into a working example: it issues a raw compound CREATE(SMB2_WRITE_DACL) + SET_INFO + CLOSE to grant "Everyone" an ACCESS_ALLOWED ACE on a file, since setting a DACL requires WRITE_DACL access that the high-level smb2_open() API has no way to request.

Add tests/prog_setsd.c + tests/test_0500_setsd_basic.sh, which round-trips a DACL through SET_INFO and QUERY_INFO against a live server to confirm the new encoder actually works end to end.

This feature was mainly tested against Samba server deployed on a Linux machine.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -17,7 +17,8 @@ set(SOURCES smb2-cat-async
             smb2-truncate-sync
             smb2-CMD-FIND
             smb2-server-sync
-            smb2-notify)
+            smb2-notify
+            smb2-set-security-info)
 
 foreach(TARGET ${SOURCES})
   add_executable(${TARGET} ${TARGET}.c)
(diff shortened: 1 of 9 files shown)


sahlberg:
Merged, thanks!

sahlberg:
Next time, use "git commit -s"   
That way you get a Sign-off line in the commit message.

RidaShamasneh:
Sure. Thanks.

PR #461, opened 2026-07-09
Add SMB2 server-side copy (FSCTL_SRV_COPYCHUNK) support

RidaShamasneh:
Adds client support for SMB2 server-side copy: request a resume key for an open source file (FSCTL_SRV_REQUEST_RESUME_KEY), then copy chunks into an open destination file (FSCTL_SRV_COPYCHUNK / _WRITE). New API: smb2_request_resume_key(), smb2_copychunk(), and a smb2_server_side_copy() convenience wrapper that does both steps in one call, each with sync and async variants.

Two correctness bugs were found and fixed while wiring up the IOCTL reply path for these ctl codes:

- struct smb2_ioctl_reply was left partially uninitialized when a reply carried no output body, so smb2_copychunk()/smb2_server_side_copy() could memcpy/smb2_free_data() a garbage pointer on any copychunk error that didn't come with limit info attached.

- FSCTL_SRV_COPYCHUNK{,_WRITE} report the chunk-count/size limit-exceeded error as STATUS_INVALID_PARAMETER but still shape the reply as a normal IOCTL response (limits in the output buffer), not the generic SMB2 error format. The reply dispatcher routed it to the generic error parser regardless, desyncing the read. The request's ctl_code is now retained on the pdu (mirroring the existing info_type/file_info_class fields kept for QUERY_INFO) so the dispatcher can tell the two reply formats apart.

Also includes:
- smb2-server-side-copy-sync example and prog_ssc test program, covering sync/async and direct-copychunk/server-side-copy-helper paths for both FSCTL variants
- negative tests that provoke real server-side rejections (chunk count over the server's limit, overlapping source/destination ranges) rather than only client-side argument validation
- reuse a single resume key across all batches of a copy instead of re-requesting one per batch

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -9,6 +9,7 @@ set(SOURCES smb2-cat-async
             smb2-raw-stat-async
             smb2-raw-getsd-async
             smb2-readlink
+            smb2-server-side-copy-sync
             smb2-lsa-lookupsids
             smb2-lseek-sync
             smb2-share-enum
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -8,6 +8,7 @@ noinst_PROGRAMS = dcerpc smb2-cat-async smb2-cat-sync \
 	smb2-raw-getsd-async \
 	smb2-raw-stat-async \
 	smb2-readlink \
+	smb2-server-side-copy-sync \
 	smb2-lsa-lookupsids \
 	smb2-lseek-sync \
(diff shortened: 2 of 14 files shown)


sahlberg:
Merged. Thanks!

RidaShamasneh:
@sahlberg This feature was tested against Samba version == 4.15.13-Ubuntu, Windows server 2016 and Windows server 2019, and other less popular SMB servers as well.

PR #464, opened 2026-07-11
Add query support and tests for file disposition info

RidaShamasneh:


RidaShamasneh:
Closing this PR:

Even though this call worked against Samba; but it failed against Windows server 2019 (i.e. server returned STATUS_INVALID_INFO_CLASS)

image

image

FileDispositionInformation is marked as Set in MS-FSCC Section 2.4, so it is not a valid information class for QUERY_INFO
image

PR #468, opened 2026-07-21
Propagate SMB domain from parsed URLs

RidaShamasneh:
In many examples, test programs, and utilities, the code already parsed SMB URLs of the form smb://[domain;][user@]host/share/path, but it only passed url->user into the SMB context before connecting. This commit adds the missing step:
```
  if (url->domain) {
      smb2_set_domain(smb2, url->domain);
  }
```
Prior to this change, the parser extracted the domain, but most callers simply discarded it. This PR also handles the case where an empty domain "" is passed in the URL, e.g.:
```
smb://\;[user@]host/share/path
"smb://;[user@]host/share/path"
```

### DIFF
--- a/examples/dcerpc.c
+++ b/examples/dcerpc.c
@@ -202,6 +202,9 @@ int main(int argc, char *argv[])
         if (url->user) {
                 smb2_set_user(smb2, url->user);
         }
+        if (url->domain) {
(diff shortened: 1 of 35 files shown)


sahlberg:
Merged. Thanks!

PR #469, opened 2026-07-24
Add SMB2 support for creating links, with an example

danbrakeley:
I needed the ability to create hard links, so I added it, including a new `examples` file for testing.

I've tested this against my Synology NAS, and it works great. I've incorporated these changes into a utility I'm working on that let's you browse/search/manage hard links on an SMB share.


sahlberg:
Nice!
Merged, thanks!

PR #471, opened 2026-07-28
Bound attacker-controlled label/name lengths in FILESYSTEM_INFO decoders

glitchfox:
As discussed by email — submitting as a PR so the credit lands in the commit.

**Bug (CWE-125, client-side, malicious/compromised SMB server):**
`smb2_decode_file_fs_volume_info()` and `smb2_decode_file_fs_attribute_info()`
(`lib/smb2-data-filesystem-info.c`) read a 32-bit label/name length from a QUERY
INFO (filesystem) reply and feed `length / 2` to `smb2_utf16_to_utf8()` without
bounding it against `vec->len`, so a short reply over-reads the heap.

**Fix:** reject replies shorter than the fixed header, and bound the label/name
length against the received reply before decoding.

**Verification:** built current master with an AddressSanitizer-instrumented
libsmb2 and the crafted volume-info / attribute-info replies — heap-buffer-overflow
before this change, clean after; valid replies still decode.

Reported privately on 2026-07-27. No CVE requested.

### DIFF
--- a/lib/smb2-data-filesystem-info.c
+++ b/lib/smb2-data-filesystem-info.c
@@ -62,12 +62,21 @@ smb2_decode_file_fs_volume_info(struct smb2_context *smb2,
         uint64_t t;
         const char *name;
 
+        if (vec->len < 18) {
+                return -1;
+        }
+
         smb2_get_uint64(vec,  0, &t);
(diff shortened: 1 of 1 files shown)


sahlberg:
Merged. Thanks

PR #472, opened 2026-08-09
dcerpc: fix building on OpenBSD

brad0:
```
libdcerpc/dcerpc-epm.c:1327:31: error: use of undeclared identifier 'AF_INET'
 1327 |                 if (inet_pton(AF_INET, ipv4, &addr) != 1) {
      |                               ^~~~~~~
libdcerpc/dcerpc-epm.c:1496:23: error: use of undeclared identifier 'AF_INET'
 1496 |         if (inet_ntop(AF_INET, rhs, ipbuf, sizeof(ipbuf)) == NULL) {
      |                       ^~~~~~~
2 errors generated.
```


sahlberg:
Merged, thanks!

PR #478, opened 2026-08-28
smb3-seal: byte-swap SessionId for the wire in smb3_encrypt_pdu()

### LINKED ISSUE #477: SMB3 encryption (seal) fails on big-endian hosts: SessionId not byte-swapped in transform header
## Body

`smb3_encrypt_pdu()` in `lib/smb3-seal.c` writes the SMB2_TRANSFORM_HEADER's `SessionId`
field with a raw `memcpy` of the host-order value. The other multi-byte fields in the same
header are explicitly converted first:

```c
u32 = htole32(spl - 52);
memcpy(&pdu->crypt[36], &u32, 4);
u16 = htole16(SMB_ENCRYPTION_AES128_CCM);
memcpy(&pdu->crypt[42], &u16, 2);
memcpy(&pdu->crypt[44], &smb2->session_id, 8);   // no endian conversion here
```

`smb2->session_id` is host byte order elsewhere in the library (`smb2_set_uint64`/
`smb2_get_uint64` in `lib/pdu.c` both go through `htole64`/`letoh64`), so this line needs
the same treatment before it hits the wire.

On little-endian hosts this is a no-op bug, since host order already equals wire order.
On big-endian it puts a byte-reversed SessionId in every encrypted PDU, which a server has
no reason to accept.

### Reproduced on hardware

PowerPC Mac (iBook G4, Tiger 10.4.11, big-endian). With `smb2_set_seal(ctx, 1)`, NEGOTIATE
and SESSION_SETUP succeed, then the connection is dropped right after the client sends its
first TRANSFORM_HEADER-wrapped PDU (TREE_CONNECT). Packet capture shows this happening
identically against two unrelated server implementations -- macOS and Windows 11 standard
SMB3 sharing (macOS closes with FIN, Windows with RST, same point in both) -- which is why
I'm fairly confident this is on the wire, not two servers sharing a bug.

### Fix

```c
{
        uint64_t sid_le = htole64(smb2->session_id);
        memcpy(&pdu->crypt[44], &sid_le, 8);
}
```

Tested against v6.0.0 with this patch: `smb2_set_seal(ctx, 1)` connects successfully to
both macOS's and Windows 11's standard SMB3 sharing.

```diff
--- a/lib/smb3-seal.c
+++ b/lib/smb3-seal.c
@@ -103,7 +103,10 @@
         memcpy(&pdu->crypt[36], &u32, 4);
         u16 = htole16(SMB_ENCRYPTION_AES128_CCM);
         memcpy(&pdu->crypt[42], &u16, 2);
-        memcpy(&pdu->crypt[44], &smb2->session_id, 8);
+        {
+                uint64_t sid_le = htole64(smb2->session_id);
+                memcpy(&pdu->crypt[44], &sid_le, 8);
+        }

         spl = 52;  /* transform header */
         for (tmp_pdu = pdu; tmp_pdu; tmp_pdu = tmp_pdu->next_compound) {
```

Can turn this into a PR if useful -- opening as an issue first since I'm new to the codebase.

### Environment
- libsmb2 v6.0.0 tag; also present on master as of 2026-08
- Client: PowerPC, Mac OS X 10.4.11 (Tiger), gcc 4.0.1, big-endian
- Servers: macOS and Windows 11 standard SMB3 file sharing, both little-endian

watermark-hd:
Fixes #477.

Per your feedback there -- switched from the scoped-variable version to a direct pointer-cast assignment:

```c
*(uint64_t *)(void *)&pdu->crypt[44] = htole64(smb2->session_id);
```

Re-tested on the same PPC hardware (Tiger iBook G4) with this exact form: `smb2_set_seal(ctx, 1)` connects successfully against both macOS's and Windows 11's standard SMB3 sharing.

### DIFF
--- a/lib/smb3-seal.c
+++ b/lib/smb3-seal.c
@@ -106,7 +106,7 @@ smb3_encrypt_pdu(struct smb2_context *smb2,
         memcpy(&pdu->crypt[36], &u32, 4);
         u16 = htole16(SMB_ENCRYPTION_AES128_CCM);
         memcpy(&pdu->crypt[42], &u16, 2);
-        memcpy(&pdu->crypt[44], &smb2->session_id, 8);
+        *(uint64_t *)(void *)&pdu->crypt[44] = htole64(smb2->session_id);
 
         spl = 52;  /* transform header */
         for (tmp_pdu = pdu; tmp_pdu; tmp_pdu = tmp_pdu->next_compound) {


sahlberg:
Awesome, thanks.  Merged!

I will set up a Linux PPC environment under qemu. I should be able to reproduce the issue there and then I can try to wire it into my test environmet.

Thanks!

PR #481, opened 2026-09-06
Fixes for macOS (older bug, new regression after 6.2)

barracuda156:
1. Fix regression due to unconditional usage of possibly non-existing CommonCrypto umbrella header.
2. Use GSS framework when it exists, and faithfully link to it, use `kerberos5` otherwise.
3. Fix a minor issue (failure with `-Werror` with clang).


barracuda156:
@sahlberg Could you please take a look when you have time?

PR #483, opened 2026-09-14
Fix/sync open and notify change leaks

damien78:
Two memory leaks found with LeakSanitizer while running an SMB2 client
test suite against a local Samba server. Both are small per-operation
leaks, but both are unbounded over a process lifetime: one is per
`smb2_open()`, the other per CHANGE_NOTIFY watch torn down.

Split into two commits so they can be taken independently.

### 1. `smb2_open()` leaks its `sync_cb_data` (16 bytes per open)

`3818a74` passed `free` as the `free_cb` argument and stored it in
`pdu->free_cb`, so `smb2_free_pdu()` released the sync wrapper's
`cb_data`. `daa92d9` removed `pdu->free_cb = free_cb;` from
`_smb2_open_async_with_oplock_or_lease()`.

That removal looks correct on its own terms: `pdu->free_cb` is invoked
as `pdu->free_cb(pdu->cb_data)`, and in the open path `pdu->cb_data` is
the `struct smb2fh`, not the caller's `cb_data` — so leaving it in
place would have freed the file handle out from under the caller.
But nothing took over freeing the sync wrapper's own `cb_data`, and the
`/* pdu takes ownership of cb_data ... */` comment above the call has
been stale since.

Fixed by freeing it in `smb2_open()` itself, which is what all the other
synchronous wrappers in `sync.c` already do, and passing `NULL` for
`free_cb` so the intent is unambiguous if the plumbing is ever restored.

### 2. `notify_change_cb()` leaks the directory handle on the cancel path

`smb2_notify_change_async()` opens a directory handle internally and
never returns it, so only libsmb2 can free it. `notify_change_cb()`
handles both normal outcomes — re-arm with the same handle when `loop`
is set, `smb2_close()` when it is not — but the error/cancellation
branch returns without either, leaking the `struct smb2fh` and its
`strdup`'d path.

That branch is precisely the teardown path (`smb2_destroy_context()`
flushing a pending CHANGE_NOTIFY with `command_data == NULL`), so every
watch that is ever torn down leaks.

It can't use `smb2_close()` there — that is the synchronous wrapper, and
this runs inside the destroy/flush where the socket is going away and
the round trip cannot complete. `free_smb2fh()` is what's wanted: the
handle's memory back, no CLOSE on the wire.

### One thing that may deserve a separate look

This is only reachable because `smb2_destroy_context()` no longer frees
outstanding handles — `daa92d9` removed the `smb2->fhs` list that used
to track them. `libsmb2.h` still documents the old contract directly
above `smb2_destroy_context()`:

> Any open "struct smb2fh" will automatically be freed. You can not
> reference any "struct smb2fh" after the context is destroyed.

That is no longer true. I have not touched it here since it is a policy
question rather than a leak, but applications written against that
sentence are leaking every handle they hold at destroy time.

### Testing

Built on top of `b3d560c` and run against a local Samba server with
ASan/LSan: 59 SMB test cases covering open/read/write/unlink, share
enumeration, directory enumeration, and CHANGE_NOTIFY (including watch
re-arm after a server bounce). Before: 422 bytes leaked in 12
allocations. After: zero. No test behaviour changed.


sahlberg:
Merged and slightly tweaked and pushed to master.

Thanks!

PR #485, opened 2026-09-20
update gamecube/wii support

dborth:
add missing compatibility shims - libogc's net_ functions don't follow posix.
update cmake so it builds properly - different detection on Windows and Linux, and with/without libwiisocket shims.


sahlberg:
Gamecube and Wii builds fail. Please update and I will merge.

